PR #1, opened 2019-04-04
fix for multiple finalizers

luthermonson:


pkg/generic/remove.go
```diff
@@ -40,12 +40,6 @@ func (o *objectLifecycleAdapter) sync(key string, obj runtime.Object) (runtime.O
 		return o.addFinalizer(obj)
 	}
 
-	finalizerKey := o.constructFinalizerKey()
-	finalizers := metadata.GetFinalizers()
-	if len(finalizers) > 0 && finalizers[0] == finalizerKey {
-		return obj, nil
-	}
-
```
ibuildthecloud: You need to check that the finalizer is in the list and then you can run `o.handler(key, obj)`

PR #40, opened 2019-09-19
Applying cluster-scoped objects with namespace metadata reveals unexpected behavior

matlec:
**Describe the bug**
While deploying the [local path provisioner](https://github.com/rancher/local-path-provisioner/blob/master/deploy/local-path-storage.yaml) with k3s using the addon controller / static manifests I encountered the following behavior:
- when the addon controller discovers and applies the manifest for the first time, the deployment succeeds as expected and all objects are successfully created on the api server
- when the addon controller tries to apply the manifest a second time (e.g. when the addon object is deleted, when the manifest changes, ...), objects that do not have namespace scope (e.g. `ClusterRole` or `ClusterRoleBinding`) but reference a namespace in their metadata (e.g. `local-path-provisioner-role` or `local-path-provisioner-bind`) are handled incorrectly. Since those objects are persisted without namespace property, [compareSets](https://github.com/rancher/wrangler/blob/master/pkg/apply/desiredset_process.go#L205) comes to the conclusion that the existing object (with stripped namespace) should be deleted and a new object should be created. [`process`](https://github.com/rancher/wrangler/blob/4a14db5d5daf08a7421c90f9e2015885f376e2c0/pkg/apply/desiredset_process.go#L149) creates objects before it deletes them. This effectively causes non-namespaced objects to be missing.

**Proposed fix**
The namespace property of objects that are not scoped by namespace can be safely removed before they are being applied. This helps [compareSets](https://github.com/rancher/wrangler/blob/master/pkg/apply/desiredset_process.go#L205) to identify existing resources and prevent them from being deleted.

### DIFF
--- a/pkg/apply/desiredset_process.go
+++ b/pkg/apply/desiredset_process.go
@@ -65,6 +65,28 @@ func (o *desiredSet) adjustNamespace(gvk schema.GroupVersionKind, objs map[objec
 	return nil
 }
 
+func (o *desiredSet) clearNamespace(objs map[objectset.ObjectKey]runtime.Object) error {
+	for k, v := range objs {
+		if k.Namespace == "" {
+			continue
+		}
+
+		v = v.DeepCopyObject()
+		meta, err := meta.Accessor(v)
+		if err != nil {
+			return err
+		}
+
+		meta.SetNamespace("")
+
+		delete(objs, k)
+		k.Namespace = ""
+		objs[k] = v
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


matlec:
As a side note, the [local path provisioner manifest](https://github.com/rancher/local-path-provisioner/blob/master/deploy/local-path-storage.yaml) should probably be fixed and have the namespace removed from cluster-wide objects :-)

ibuildthecloud:
Cool, thanks!

PR #45, opened 2019-10-30
Add WithGVK() method to Apply interface

daxmc99:
This new method is implemented in desiredSet to pass in a known
set of gvks to manage previously created resources.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -49,6 +49,7 @@ type Apply interface {
 	WithListerNamespace(ns string) Apply
 	WithRateLimiting(ratelimitingQps float32) Apply
 	WithNoDelete() Apply
+	WithGVK(gvks ...schema.GroupVersionKind) Apply
 	WithSetOwnerReference(controller, block bool) Apply
 }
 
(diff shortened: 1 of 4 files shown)


pkg/apply/apply.go
```diff
@@ -170,6 +171,10 @@ func (a *apply) WithCacheTypes(igs ...InformerGetter) Apply {
 	return a.newDesiredSet().WithCacheTypes(igs...)
 }
 
+func (a *apply) WithGVK(gvks []schema.GroupVersionKind) Apply {
```
ibuildthecloud: I'd prefer that be 
```suggestion
func (a *apply) WithGVK(gvks ..,schema.GroupVersionKind) Apply {
```

pkg/gvk/get.go
```diff
@@ -45,3 +44,18 @@ func Set(obj runtime.Object) error {
 	kind.SetGroupVersionKind(gvks[0])
 	return nil
 }
+
+func ConvertObjs(objs []runtime.Object) ([]schema.GroupVersionKind, error) {
```
ibuildthecloud: What is this method for

pkg/apply/desiredset.go
```diff
@@ -56,6 +56,17 @@ func (o desiredSet) ApplyObjects(objs ...runtime.Object) error {
 	return o.Apply(os)
 }
 
+// WithGVK uses a known listing of existing gvks to modify the the prune types to allow for deletion of objects
+func (o desiredSet) WithGVK(gvks []schema.GroupVersionKind) Apply {
+	if o.pruneTypes == nil {
```
ibuildthecloud: WithCacheTypes will blank out this map, so you need to edit that method too.  Also this method provides a clone of the object so if pruneType != nil you need to copy the map to a new map and then set the new gvks.

pkg/gvk/get.go
daxmc99: https://github.com/daxmc99/rio/blob/f6f1807b1e27b645519201918e96d4b4b5ec40d6/cli/cmd/up/up.go#L153 
Helper func for handling lists of objects, this will likely go away with a change to variadic functions.

PR #47, opened 2019-11-08
Fix onwer reference bug

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/726

PR #50, opened 2019-12-05
fix status handler conditions being always empty

### LINKED ISSUE #49: generated status handler processing never updates conditions
While working on the upgrade operator for k3os I've noticed that conditions are never updated and propagated even though the handler has mutated the status.

dweomer:
Fixes #49 that prevented status updates from accumulating conditions.

### DIFF
--- a/pkg/controller-gen/generators/type_go.go
+++ b/pkg/controller-gen/generators/type_go.go
@@ -360,6 +360,7 @@ func (a *{{.lowerName}}StatusHandler) sync(key string, obj *{{.version}}.{{.type
 	}
 	if !equality.Semantic.DeepEqual(status, newStatus) {
 		var newErr error
+		newStatus.Conditions = obj.Status.Conditions
 		obj.Status = newStatus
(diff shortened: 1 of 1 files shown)


pkg/controller-gen/generators/type_go.go
```diff
@@ -360,6 +360,7 @@ func (a *{{.lowerName}}StatusHandler) sync(key string, obj *{{.version}}.{{.type
 	}
 	if !equality.Semantic.DeepEqual(status, newStatus) {
 		var newErr error
+		newStatus.Conditions = obj.Status.Conditions
```
dweomer: pickup the results of the `a.condition.SetError` operations from above

daxmc99: Hmm, it looks likes newStatus never escapes this scope.

dweomer: it's copied to `obj.Status` and then persisted via `UpdateStatus` on `obj`

dweomer:
Superceded by #51

PR #59, opened 2020-02-02
WIP: WithoutOwnerReference

cbron:
For deployment wrangler


dweomer:
@cbron is the use case here so that you can return create an object that wont get deleted with it's parent goes away and/or doesn't get returned on on subsequent invocations to the generating handler? I thought `WithNoDelete` covered these.

cbron:
@dweomer The first part is our use case, but I don't believe `WithNoDelete` works for that. That seems to only come into play when delete is part of an Apply chain via wrangler. In this case it appears the resource (a deployment) is getting cleaned up by Kubernetes garbage collection because of `ownerReferences` annotation.

cbron:
Closing in favor of https://github.com/rancher/wrangler/pull/60 which is object specific.

PR #60, opened 2020-02-04
WIP: Allow disabling ownerRef via annotation

cbron:
Changed meta name to vMeta so it didn't clash with import.
Using rio.cattle ns because everything else does 🤷‍♂


cbron:
Closing, likely no longer needed.

PR #66, opened 2020-02-22
fix spurious replace wait errors

dweomer:
This address an issue in rancher/system-upgrade-controller#25 when generating objects `WithNoDelete()` where the replace logic fails with noisy 'Replace Wait' errors in the log.
This also mitigates the no-delete+replace logic swamping the rate limit for a custom type.

### DIFF
--- a/pkg/apply/desiredset_compare.go
+++ b/pkg/apply/desiredset_compare.go
@@ -58,7 +58,7 @@ func prepareObjectForCreate(gvk schema.GroupVersionKind, obj runtime.Object) (ru
 }
 
 func originalAndModified(gvk schema.GroupVersionKind, oldMetadata v1.Object, newObject runtime.Object) ([]byte, []byte, error) {
-	original, err := getOriginal(gvk, oldMetadata)
+	original, err := getOriginalBytes(gvk, oldMetadata)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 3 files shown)


dweomer:
This maintains the previous `apply.WithNoDelete()` behavior while marking the option as deprecated and pointing to the newly introduced `apply.WithDeletePolicy()`.

dweomer:
@ibuildthecloud I tried modifying the `apply.Reconciler` signature to pass down the patch type and bytes throwing `apply.ErrReplace` when `.spec.template` is present, as you suggested, but it does not mitigate the use-case of generating jobs with some/any of the `PodTemplateSpec` default values missing. It could be the use of DownwardAPI volume mounts is gumming it up but so long as I generate `Jobs` with all of their defaults the compare/reconciler logic doesn't earn me spurious replacements.

Here is an example of the type of `job.spec.template` patches snippets that I was seeing when generating jobs without defaults:
```json
{
  "spec": {
    "/volumes": [
      {
        "name": "host-root"
      },
      {
        "name": "pod-info"
      }
    ],
    "volumes": [
      {
        "": [
          "downwardAPI",
          "name"
        ],
        "downwardAPI": {
          "items": [
            {
              "fieldRef": {
                "fieldPath": "metadata.labels"
              },
              "path": "labels"
            },
            {
              "fieldRef": {
                "fieldPath": "metadata.annotations"
              },
              "path": "annotations"
            }
          ]
        },
        "name": "pod-info"
      }
    ]
  }
}
```

dweomer:
@ibuildthecloud my latest push has this working without the need to set defaults when generating jobs

PR #70, opened 2020-03-10
Add error summarizer

brendarearden:
Adding a new error types to be used for catalogs

**Issue:**
https://github.com/rancher/rancher/issues/24914

### DIFF
--- a/pkg/summary/summarizers.go
+++ b/pkg/summary/summarizers.go
@@ -50,6 +50,7 @@ var (
 		"ScalingActive":               "pending",
 		"AbleToScale":                 "pending",
 		"RunCompleted":                "running",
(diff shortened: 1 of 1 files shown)


pkg/summary/summarizers.go
```diff
@@ -50,6 +50,7 @@ var (
 		"ScalingActive":               "pending",
 		"AbleToScale":                 "pending",
 		"RunCompleted":                "running",
+		"Errored": "error",
```
ibuildthecloud: Can you run gofmt?  I'd consider a different name for the condition.  False means "bad", so Errored=False is bad and Errored=True is good.  Somewhat counter intuitive.

PR #89, opened 2020-04-28
Update lastUpdatedTime for StatusHandler when Status Changes

ekristen:
I am still testing this change locally, but appears to be working as expected.


pkg/controller-gen/generators/type_go.go
```diff
@@ -368,6 +368,9 @@ func (a *{{.lowerName}}StatusHandler) sync(key string, obj *{{.version}}.{{.type
 		}
 	}
 	if !equality.Semantic.DeepEqual(origStatus, &newStatus) {
+		// Since status has changed, update the lastUpdatedTime
+		a.condition.LastUpdated(&newStatus, time.Now().UTC().Format(time.RFC3339))
```
ibuildthecloud, line 372: similar to L363 check for condition != ""

ekristen: Fixed.

PR #95, opened 2020-06-04
Drop resourceVersion on lists when set to 0

dramich:
Problem:
In a cluster with a large number of a single object, say 10k large
secrets, when the controller starts it attempts to list all of those
secrets and hits the client timeout

Solution:
Just increasing the client timeout does not work as the API server also
has a timeout. The API server ignores the limit when resourceVersion is
set to 0 which causes the call to take longer than either timeout
(running curl against the API server the call was taking over 3min on
average and pulling approx 2GB of data) so if resourceVersion is set to
0 reset it to empty string


dramich:
https://github.com/rancher/rancher/issues/26166

PR #114, opened 2020-08-25
Add git library to wrangler

StrongMonkey:


pkg/git/git.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	cmd.Env = append(cmd.Env, "SSH_AUTH_SOCK="+addr.Name)
+
+	go func() {
+		defer l.Close()
+		for {
+			conn, err := l.Accept()
+			if err != nil {
+				if !k8snet.IsProbableEOF(err) {
```
StrongMonkey, line 246: This is added so that some EOF can be avoided instead of printing in logs

pkg/git/git.go
```diff
@@ -0,0 +1,279 @@
...
+					logrus.Errorf("failed to accept ssh-agent client connection: %v", err)
+				}
+				return
+			}
+			if err := agent.ServeAgent(*g.agent, conn); err != nil && k8snet.IsProbableEOF(err) {
```
StrongMonkey, line 251: same

StrongMonkey:
This library is taken directly from https://github.com/rancher/rancher/blob/master/pkg/catalogv2/git/git.go#L272 but removed remoteSHAchanged logic since it requires custom github logic.

PR #132, opened 2020-10-03
Fix ssh agent to use tmpfile

StrongMonkey:


pkg/git/git.go
```diff
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-
```
ibuildthecloud, line 21: You have been trained well :)

StrongMonkey: yes yes!

StrongMonkey: this is called `reflex arc`. When I saw an empty line in import I have reflex arc to remove it :)

PR #135, opened 2020-10-11
Fix http password authentication for git

### LINKED ISSUE #134: Git http password authentication don't work
Git access to repository with http basic authentication don't work with the following error:

```
git ls-remote https://user@example.com/user/repository.git refs/heads/master error: exit status 128, detail: fatal: could not read Password for 'https://user@example.com': terminal prompts disabled
```

Because reading password from `Stdin` don't work for git command: [https://github.com/rancher/wrangler/blob/264809ad4ab939384badb15f12c548ad0fd776ca/pkg/git/git.go#L321]

A possible solution is supply password on http URL.

smoretto:
Closes #134

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -262,7 +262,7 @@ func (g *Git) setCredential(cred *corev1.Secret) error {
 		if err != nil {
 			return err
 		}
-		u.User = url.User(string(username))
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
The problem with this approach is that it puts the password in the command line, which can be observed by other programs.

PR #137, opened 2020-10-19
Fix http auth

StrongMonkey:
Use "git -c" to adopt http basic auth from env variables to avoid
reading password from URL. Pass password to stdin doesn't work as we
disabled git prompt.


StrongMonkey:
https://github.com/rancher/fleet/issues/118

PR #150, opened 2021-02-11
Fix summarizing of CNCF Rook resources

ron1:
Fixing summarizing of CNCF Rook project custom resources including CephCluster and io.objectbucket.v1alpha1.ObjectBucketClaim. This PR fixes issues https://github.com/rancher/wrangler/issues/148 and https://github.com/rancher/wrangler/issues/149.

### DIFF
--- a/pkg/summary/summarizers.go
+++ b/pkg/summary/summarizers.go
@@ -77,18 +77,16 @@ var (
 	// False == error
 	// Unknown ==
 	ErrorFalse = map[string]bool{
-		"Failed":      true,
-		"Progressing": true,
+		"Failed": true,
 	}
 
 	// True ==
 	// False == transitioning
 	// Unknown == error
 	TransitioningFalse = map[string]string{
-		"Completed":   "activating",
-		"Ready":       "unavailable",
-		"Available":   "updating",
-		"Progressing": "inactive",
+		"Completed": "activating",
+		"Ready":     "unavailable",
+		"Available": "updating",
 	}
 
 	// True ==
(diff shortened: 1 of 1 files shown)


nickgerace:
Functionally, this looks good to me. However, I will admit that me knowledge of Rook is thinner than yours. I'll run some tests on my side to ensure this change does not affect anything downstream. Wrangler is used by repos other than Fleet, for example.

ron1:
`Progressing` is a common type of condition used to provide status info for CNCF Rook CephCluster CRs and lots of other CRs. For some unknown reason, wrangler currently summarizes the `Progressing` condition as a non-Transitioning conditioning just like `Completed`, `Ready`, or `Available`. Instead it should be summarized as a Transitioning condition just like `Reconciling`. I suspect other CRs besides CephCluster will benefit from this change.

ibuildthecloud:
@ron1 I think how Rook uses Progressing is different from how Deployment uses Progressing.  For a deployment, afaik, Progressing=False is an error state and Progressing=True is normal state, not transitioning.

What I would recommend here is that Progressing just be removed as a condition we check.  The kstatus checks will handle the Deployment now so we don't need to worry about that condition.

ron1:
@ibuildthecloud Based on your feedback, I will revise my PR to remove the Progressing references. Thanks for the review.

ron1:
@ibuildthecloud @nickgerace Per your recommendation, I pushed a revised commit that simply removes Progressing as a condition that is explicitly checked. Please let me know if you have additional questions. It would be great if this PR could be incorporated into the wrangler version included with Rancher 2.5.6 and Fleet 0.3.4. I suspect any SUSE CaaS Platform users trying to use Rancher/Fleet to deploy Rook Ceph to downstream clusters will also benefit from this PR.

ron1:
@ibuildthecloud @nickgerace Thanks much for the PR review and commit!

nickgerace:
No problem. Was going to approve. @ibuildthecloud beat me to it.

Thanks for the PR again.

PR #154, opened 2021-03-12
Update crd version

StrongMonkey:


ibuildthecloud:
In order to accept this PR I think we need to change the CRD generate such that if no openapi schema is provided we just add the following schema.  Then this will continue to work with Rancher until we fix the CRDs in Rancher.
```
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            x-kubernetes-preserve-unknown-fields: true
          status:
            x-kubernetes-preserve-unknown-fields: true```

ekristen:
@ibuildthecloud once this is accepted will wrangler defaults for CRD properties work or will that require an additional PR to work?

ibuildthecloud:
@ekristen Defaults should work.

StrongMonkey:
@ibuildthecloud Dummy schema is added to prevent pruning.

PR #156, opened 2021-04-06
Use crypto/ssh and update library to support more key format

StrongMonkey:


StrongMonkey:
https://github.com/rancher/fleet/issues/221

PR #160, opened 2021-04-22
adding coordination/v1 to generated controllers

ekristen:
This got added as `coordination.k8s.io` vs `coordination` I see both are being done, is there a preference for one vs the other?


StrongMonkey (approved):

ibuildthecloud:
The full coordination.k8s.io form is preferred.  I think the shorter ones are for legacy reasons.

PR #166, opened 2021-06-21
Removed v1beta1 Ingress reference

snasovich:
v1beta1 Ingress is getting removed in k8s v1.22.
Usages of Ingress defined in Wrangler were not found in rancher org codebase (https://github.com/search?q=org%3Arancher+wrangler%2Fpkg%2Fgenerated%2Fcontrollers%2Fextensions&type=code) and thus it's suggested to proactively remove it.

Please let me know if I've missed some usage of it.


snasovich:
This is misc change for the following issue: https://github.com/rancher/rancher/issues/33312

dramich:
Can you split into two commits, one for your change and another for generated changes

snasovich:
> Can you split into two commits, one for your change and another for generated changes

@dramich , done.

dramich (approved):

snasovich:
Closing as we may need to implement conditional logic for #33312 and this change doesn't make sense then.

PR #169, opened 2021-07-02
Add debug logging around webhook handler invocation

ryansann:
These changes add access logs to webhook handlers with a middleware pattern. The logs have format that looks like:
```
DEBU[0003] CREATE management.cattle.io/v3.GlobalRole gr-wgmzb user=user-j9rfg allowed=false err=<nil> 
```
To enable, applications using the webhook package need to set logrus to the debug level

### DIFF
--- a/pkg/webhook/router.go
+++ b/pkg/webhook/router.go
@@ -74,13 +74,16 @@ func (r *Router) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 func (r *Router) admit(response *Response, request *v1.AdmissionRequest, req *http.Request) error {
 	for _, m := range r.matches {
 		if m.matches(request) {
-			return m.admit(response, &Request{
+			err := m.admit(response, &Request{
 				AdmissionRequest: *request,
 				Context:          req.Context(),
 				objTemplate:      m.getObjType(),
 			})
+			logrus.Debugf("admit result: %s %s %s user=%s allowed=%v err=%v", request.Operation, request.Kind.String(), resourceString(request.Namespace, request.Name), request.UserInfo.Username, response.Allowed, err)
+			return err
 		}
 	}
+	logrus.Debugf("no route match found for %s %s %s", request.Operation, request.Kind.String(), resourceString(request.Namespace, request.Name))
 	return nil
 }
 
@@ -141,4 +144,12 @@ type HandlerFunc func(resp *Response, req *Request) error
 
 func (h HandlerFunc) Admit(resp *Response, req *Request) error {
 	return h(resp, req)
+}
+
+// resourceString returns the resource formatted as a string
+func resourceString(ns, name string) string {
+	if ns == "" {
+		return name
+	}
+	return fmt.Sprintf("%s/%s", ns, name)
 }


pkg/webhook/middleware.go
```diff
@@ -0,0 +1,27 @@
...
+
+func access(next Handler) Handler {
+	return HandlerFunc(func(res *Response, req *Request) error {
+		err := next.Admit(res, req)
+		logrus.Debugf("%s %s %s user=%s allowed=%v err=%v", req.Operation, gvk(req.Kind), resource(req.Namespace, req.Name), req.UserInfo.Username, res.Allowed, err)
+		return err
+	})
+}
+
+func gvk(k v1.GroupVersionKind) string {
```
dramich, line 18: Update the function name to clarify what it does and add a comment

pkg/webhook/middleware.go
```diff
@@ -0,0 +1,27 @@
...
+	return fmt.Sprintf("%s/%s.%s", k.Group, k.Version, k.Kind)
+}
+
+func resource(ns, name string) string {
```
dramich, line 22: Update the function name to clarify what it does and add a comment

pkg/webhook/router.go
```diff
@@ -81,6 +81,7 @@ func (r *Router) admit(response *Response, request *v1.AdmissionRequest, req *ht
 			})
 		}
 	}
+	logrus.Debugf("no route match found for %s %s %s", request.Operation, gvk(request.Kind), resource(request.Namespace, request.Name))
```
dramich, line 84: Nice

pkg/webhook/match.go
```diff
@@ -74,10 +74,13 @@ func checkBool(expected, actual *bool) bool {
...
-func (r *RouteMatch) DryRun(dryRun bool) *RouteMatch               { r.dryRun = &dryRun; return r }
-func (r *RouteMatch) Group(group string) *RouteMatch               { r.group = group; return r }
-func (r *RouteMatch) HandleFunc(handler HandlerFunc) *RouteMatch   { r.handler = handler; return r }
-func (r *RouteMatch) Handle(handler Handler) *RouteMatch           { r.handler = handler; return r }
+func (r *RouteMatch) DryRun(dryRun bool) *RouteMatch { r.dryRun = &dryRun; return r }
+func (r *RouteMatch) Group(group string) *RouteMatch { r.group = group; return r }
+func (r *RouteMatch) HandleFunc(handler HandlerFunc) *RouteMatch {
+	r.handler = access(handler)
+	return r
+}
```
dramich, line 82: This isn't needed, catch the admit error and log based off of that.

pkg/webhook/router.go
```diff
@@ -74,13 +74,16 @@ func (r *Router) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 func (r *Router) admit(response *Response, request *v1.AdmissionRequest, req *http.Request) error {
 	for _, m := range r.matches {
 		if m.matches(request) {
-			return m.admit(response, &Request{
+			err := m.admit(response, &Request{
 				AdmissionRequest: *request,
 				Context:          req.Context(),
 				objTemplate:      m.getObjType(),
 			})
+			logrus.Debugf("%s %s %s user=%s allowed=%v err=%v", request.Operation, request.Kind.String(), resourceString(request.Namespace, request.Name), request.UserInfo.Username, response.Allowed, err)
```
dramich, line 82: Prefix the log message so we can see where it's coming from and have some context

dramich (approved):

PR #173, opened 2021-07-20
Add noDeleteGVK

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/33203

This PR would ensure that certain GVKs are not deleted for desiredset applys.

### DIFF
--- a/pkg/apply/desiredset.go
+++ b/pkg/apply/desiredset.go
@@ -36,6 +36,7 @@ type desiredSet struct {
 	informerFactory          InformerFactory
 	remove                   bool
 	noDelete                 bool
+	noDeleteGVK              map[schema.GroupVersionKind]struct{}
 	setID                    string
 	objs                     *objectset.ObjectSet
 	codeVersion              string
@@ -222,6 +223,16 @@ func (o desiredSet) WithRateLimiting(ratelimitingQps float32) Apply {
 
 func (o desiredSet) WithNoDelete() Apply {
 	o.noDelete = true
+	return o
+}
+
+func (o desiredSet) WithNoDeleteGVK(gvks ...schema.GroupVersionKind) Apply {
+	if o.noDeleteGVK == nil {
+		o.noDeleteGVK = make(map[schema.GroupVersionKind]struct{})
+	}
+	for _, curr := range gvks {
+		o.noDeleteGVK[curr] = struct{}{}
+	}
 	return o
 }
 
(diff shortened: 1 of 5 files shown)


pkg/apply/desiredset.go
```diff
@@ -36,6 +36,7 @@ type desiredSet struct {
 	informerFactory          InformerFactory
 	remove                   bool
 	noDelete                 bool
+	noDeleteGVK              []schema.GroupVersionKind
```
StrongMonkey, line 39: This can be map[schema.GroupVersionKind] so you can do fast lookup and avoid duplicate entry. Overall approaches looks good

nickgerace: Ok I will change to that. Add a dummy bool for the value then?

StrongMonkey: yeah dummy struct{} is probably ok.

pkg/apply/desiredset_crud.go
```diff
@@ -51,9 +52,16 @@ func (o *desiredSet) get(nsed bool, namespace, name string, client dynamic.Names
...
 
-func (o *desiredSet) delete(nsed bool, namespace, name string, client dynamic.NamespaceableResourceInterface, force bool) error {
-	if o.noDelete && !force {
-		return nil
+func (o *desiredSet) delete(nsed bool, namespace, name string, client dynamic.NamespaceableResourceInterface, force bool, gvk schema.GroupVersionKind) error {
+	if !force {
+		if o.noDelete {
+			return nil
+		}
+		for key := range o.noDeleteGVK {
```
StrongMonkey, line 60: You can just do 
if _, ok := o.noDeleteGVK[key]; ok {
  return nil
}

pkg/apply/desiredset.go
```diff
@@ -225,6 +226,13 @@ func (o desiredSet) WithNoDelete() Apply {
 	return o
 }
 
+func (o desiredSet) WithNoDeleteGVK(gvk ...schema.GroupVersionKind) Apply {
+	for _, curr := range gvk {
+		o.noDeleteGVK[curr] = struct{}{}
```
StrongMonkey, line 231: I believe you have to initialize the map first.

pkg/apply/desiredset_crud.go
nickgerace: _facepalm_ thank you. That's what I get for quick refactoring.

pkg/apply/desiredset.go
nickgerace: Still drinking coffee over here, thank you.

StrongMonkey (approved):

dramich (approved):

PR #174, opened 2021-07-29
fix goroutine leaks, when using 'for range ticker.Context()' in gorou…

watjt:
"ticker.Context()" used in many places in the rancher,  like this "for range ticker.Context() {do something}", the channel will not be closed for "ticker.Stop()", "for range  ticker.Context() {do something}" outside the goroutine will never exit, cause goroutine leaks.

### DIFF
--- a/pkg/ticker/ticker.go
+++ b/pkg/ticker/ticker.go
@@ -7,9 +7,18 @@ import (
 
 func Context(ctx context.Context, duration time.Duration) <-chan time.Time {
 	ticker := time.NewTicker(duration)
+	c := make(chan time.Time)
 	go func() {
-		<-ctx.Done()
-		ticker.Stop()
+		for {
+			select {
+			case t := <-ticker.C:
+				c <- t
+			case <-ctx.Done():
+				close(c)
+				ticker.Stop()
+				return
+			}
+		}
 	}()
-	return ticker.C
+	return c
 }


dramich:
@watjt Thanks for the pr and bringing this up. I think this only solves part of the problem as the ticker channel will still not get closed and will still leak even with this fix. https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/time/tick.go;l=44

watjt:
> @watjt Thanks for the pr and bringing this up. I think this only solves part of the problem as the ticker channel will still not get closed and will still leak even with this fix. https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/time/tick.go;l=44

Yes,  the ticker channel will still not get closed when goroutine where ticker is located still not exit. if we free *Ticker (Exit the goroutine where ticker is located), ticker channel will be gc. goroutine that uses' for range ticker.Context () 'will also exit. the main reason is that the goroutine here has leaked.

ibuildthecloud:
The ticker should get GC'd with this fix so the open channel won't matter.

LGTM

watjt:
> The ticker should get GC'd with this fix so the open channel won't matter.
> 
> LGTM


You see how it works in rancher.  if i create user cluster and delete, will these goroutines exit?

https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuser/healthsyncer/healthsyncer.go#L60
https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuserlegacy/alert/watcher/pod.go#L67
https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuserlegacy/alert/watcher/node.go#L46
https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuserlegacy/alert/watcher/syscomponent.go#L38
https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuserlegacy/alert/watcher/workload.go#L75
https://github.com/rancher/rancher/blob/2b156fd2b0befb0904df08621b2577e28881fd1e/pkg/controllers/managementuserlegacy/approuter/approuter.go#L27
etc...

PR #176, opened 2021-07-29
ConfigMaps from projected volumes were being reported as Secrets

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/33657

PR #178, opened 2021-08-03
add ready state to summarizer

kinarashah:
This was removed as a part of https://github.com/rancher/wrangler/commit/3eba78f45e7d61364efd11d73334641c17305b4c. Not sure if it was intentional, but this causes state to be Active for RKE1 nodes (and maybe clusters) even if the conditions have the right information. 

Issue: 
https://github.com/rancher/rancher/issues/32891

### DIFF
--- a/pkg/summary/summarizers.go
+++ b/pkg/summary/summarizers.go
@@ -86,6 +86,7 @@ var (
 	// Unknown == error
 	TransitioningFalse = map[string]string{
 		"Completed":           "activating",
+		"Ready":               "unavailable",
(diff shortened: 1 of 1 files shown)


kinarashah:
Other possible place I could think of was adding `Unknown` here along with checking for `False`
https://github.com/rancher/wrangler/blob/master/pkg/summary/summarizers.go#L284

```
		if c.Type() == "Ready" && c.Status() == "False" {
			ready = false
			readyMessage = c.Message()
			continue
		}
```

PR #180, opened 2021-10-01
feat(go.mod): update lasso version

FrankYang0529:
Update https://github.com/rancher/lasso version to fix apiextensions.k8s.io/v1beta1 CustomResourceDefinition warning message.

Related PR: https://github.com/rancher/lasso/pull/10


dramich:
@FrankYang0529 We can't merge this due to Rancher's current stance on being backwards compatible with k8s cluster versions. This could cause issues if the cluster version is pre 1.16 as there is currently no way to handle both API endpoints. I am going to close this as it's part of a larger internal discussion on how to handle this.

PR #182, opened 2021-11-03
adding the RequestShutdown func from upstream

luthermonson:
trying to get this shutdown func to use from windows service manager to cancel context when a service is stopped via Stop-Service in powershell.

source: https://github.com/kubernetes/kubernetes/blob/da707b6133ff0c66dc4d24295374e7e5de8852f5/staging/src/k8s.io/apiserver/pkg/server/signal.go#L59-L69

example usage: https://github.com/kubernetes/kubernetes/blob/da707b6133ff0c66dc4d24295374e7e5de8852f5/pkg/windows/service/service.go#L82-L87

### DIFF
--- a/pkg/signals/signal.go
+++ b/pkg/signals/signal.go
@@ -1,12 +1,9 @@
 /*
 Copyright 2017 The Kubernetes Authors.
-
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


luthermonson:
https://github.com/rancher/rke2/issues/1755

PR #183, opened 2021-11-09
fixed permissions errors when applying multiple namespaces

paynejacob:
https://github.com/rancher/rancher/issues/34712

# Problem
When users try  to modify resources in multiple namespaces wrangler attempts to list them across all namespaces.  Most users do not have permission to list resources across all namespaces and an error is returned.

# Solution
When applying resources gather what namespaces will be affected make multiple searches across the specified namespaces.

# Testing
I tested this and confirmed users will still see permission errors if they attempt to create a resource in a namespace that they do not have access to.  I also  tested creating resources that are not namespaced with a user that has access to them.

### DIFF
--- a/pkg/apply/desiredset_process.go
+++ b/pkg/apply/desiredset_process.go
@@ -1,17 +1,21 @@
 package apply
 
 import (
+	"context"
 	"fmt"
 	"sort"
+	"sync"
 
 	"github.com/pkg/errors"
 	gvk2 "github.com/rancher/wrangler/pkg/gvk"
 	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/rancher/wrangler/pkg/objectset"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 	errors2 "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -341,25 +345,21 @@ func (o *desiredSet) list(namespaced bool, informer cache.SharedIndexInformer, c
 	)
 
 	if informer == nil {
-		var c dynamic.ResourceInterface
+		// if a lister namespace is set assume all objects will belong to the listerNamespace
+		var namespaces []string
 		if o.listerNamespace != "" {
-			c = client.Namespace(o.listerNamespace)
+			namespaces = append(namespaces, o.listerNamespace)
 		} else {
-			c = client
+			namespaces = o.objs.Namespaces()
 		}
 
-		list, err := c.List(o.ctx, v1.ListOptions{
-			LabelSelector: selector.String(),
-		})
-		if err != nil {
-			return nil, err
-		}
-
-		for _, obj := range list.Items {
-			copy := obj
-			if err := addObjectToMap(objs, &copy); err != nil {
+		err := multiNamespaceList(o.ctx, namespaces, client, selector, func(obj unstructured.Unstructured) {
+			if err := addObjectToMap(objs, &obj); err != nil {
 				errs = append(errs, err)
 			}
+		})
+		if err != nil {
+			errs = append(errs, err)
 		}
 
 		return objs, merr.NewErrors(errs...)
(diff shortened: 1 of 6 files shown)


jakefhyde (approved):

pkg/apply/desiredset_process.go
```diff
@@ -433,3 +433,33 @@ func addObjectToMap(objs map[objectset.ObjectKey]runtime.Object, obj interface{}
 
 	return nil
 }
+
+// listAllByNamespace lists objects across all given namespaces, because requests are concurrent it is possible for appendFn to be called before errors are reported.
```
snasovich, line 437: Comment and func name mismatch

pkg/apply/desiredset_process.go
```diff
@@ -433,3 +433,33 @@ func addObjectToMap(objs map[objectset.ObjectKey]runtime.Object, obj interface{}
...
+			for _, obj := range list.Items {
+				appendFn(obj)
+			}
+			mu.Unlock()
+
+			return nil
+		})
+	}
+
+	// wait for all requests
```
snasovich, line 463: Nit, but I think this comment is trivial

snasovich (commented):
Did a quick review, couple minor comments. Will try to review again later.

jakefhyde (approved):

pkg/objectset/objectset.go
```diff
@@ -167,6 +167,31 @@ func (o *ObjectSet) GVKOrder(known ...schema.GroupVersionKind) []schema.GroupVer
 	return append(o.gvkOrder, rest...)
 }
 
+// Namespaces all distinct namespaces found on the objects in this set.
+func (o *ObjectSet) Namespaces() (namespaces []string) {
+	for _, _gvk := range o.ObjectsByGVK() {
```
snasovich, line 172: Any specific reason it's `_gvk` and not just `_gvk`? In fact `gvk` is probably not the best name either as GVK is key in `ObjectsByGVK` and value is `map[ObjectKey]runtime.Object` - so `objsByKey` is probably more descriptive.

pkg/objectset/objectset_test.go
```diff
@@ -0,0 +1,84 @@
...
+				errs:        tt.fields.errs,
+				objects:     tt.fields.objects,
+				objectsByGK: tt.fields.objectsByGK,
+				order:       tt.fields.order,
+				gvkOrder:    tt.fields.gvkOrder,
+				gvkSeen:     tt.fields.gvkSeen,
+			}
+
+			gotNamespaces := o.Namespaces()
+			assert.ElementsMatchf(t, tt.wantNamespaces, gotNamespaces, "Namespaces() = %v, want %v", gotNamespaces, tt.wantNamespaces)
```
snasovich, line 81: +1 for `assert.ElementsMatchf` as order of results is undetermined (map iteration order is randomized)

pkg/apply/desiredset_process.go
```diff
@@ -433,3 +433,32 @@ func addObjectToMap(objs map[objectset.ObjectKey]runtime.Object, obj interface{}
...
 }
+
+// multiNamespaceList lists objects across all given namespaces, because requests are concurrent it is possible for appendFn to be called before errors are reported.
+func multiNamespaceList(ctx context.Context, namespaces []string, baseClient dynamic.NamespaceableResourceInterface, selector labels.Selector, appendFn func(obj unstructured.Unstructured)) error {
+	var mu sync.Mutex
+	wg, _ctx := errgroup.WithContext(ctx)
+
+	// list all namespaces concurrently
+	for _, namespace := range namespaces {
+		namespace := namespace
```
snasovich, line 444: I understand we create a copy of namespace but this statement still looks weird to me. In previous code we've had `copy := obj` in the similar situation so maybe we could do `namespaceCopy := namespace` for clarity.
If this `a := a` construct is something we regularly use I'm OK with keeping it as is.

cmurphy (approved):
Glad to see tests here! Would be nice if there was also a drone job to run them automatically. I ran them locally and at least the new ones seem to work.

pkg/apply/desiredset_process_test.go
```diff
@@ -0,0 +1,93 @@
...
+	baseClient.PrependReactor("list", "*", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+		if strings.Contains(action.GetNamespace(), "error") {
+			return true, nil, errors.New("simulated failure")
+		}
+
+		return true, results[action.GetNamespace()], nil
+	})
+
+	// no namespaces
+	{
```
snasovich, line 42: In `pkg/objectset/objectset_test.go` we use "more native" approach and do table-driven tests by defining `tests` slice and iterating over it. This produces better `go test` output.
I believe it should be possible to rewrite this test with the same approach or am I missing a reason they are done this way here?

snasovich (changes requested):
Kudos for adding tests for this!
Have a couple of additional comments though, apologies for not sending them with the first review.

pkg/apply/desiredset_process.go
paynejacob: declaring it with the same name has the same effect as `copy := obj`.  I have seen this pattern elsewhere in rancher but I don't think it is standardized.  I would argue this is a safer method,  it prevents someone from absentmindedly referencing `namespace` instead of `namespaceCopy`.  That mistake can sometimes hide and produces really strange behavior that can be hard to debug later.

pkg/apply/desiredset_process_test.go
paynejacob: When I initially started theses tests it seemed like I would need to call things between cases but looking now I can use that same pattern.  Ill make them consistent.

pkg/apply/desiredset_process.go
snasovich: I agree with your rationale here. Let's keep it as is, thank you.

jakefhyde (approved):

pkg/apply/desiredset_process_test.go
```diff
@@ -0,0 +1,108 @@
...
+				calls += 1
+			})
+
+			if tt.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+
+			if tt.expectedCalls > 0 {
```
snasovich, line 103: Should be `>=` to fully validate the first test case.

snasovich (changes requested):
One trivial thing to fix

snasovich (approved):
LGTM

cmurphy (approved):

kinarashah (approved):
Great job on adding tests! 🎉 🥇

PR #186, opened 2021-11-25
fix: TrimRight removes an extra character in certain conditions

ekristen:
TrimRight isn't the right function to use here, in fact it can cause all sorts of unwanted behavior because the cutset is going to be something like `/v1` or `/v1beta1` and TrimRight will remove any of the characters regardless of order (at least how I understand it) from the right if they exist. 

We actually only want to TrimSuffix when it specifically matches `/+gv.Version` so `/v1` or `/v1beta1` etc.

I was using `project.ekristen.dev` when this code was generating `project.ekristen.de` because it was Trimming right `/v1` from `project.ekristen.dev/v1` so it was removing extra v because the cutset is `/v1`


brandond (approved):

snasovich (approved):

snasovich:
@ekristen, thank you for reporting and addressing this issue.

PR #188, opened 2021-12-01
Optimize getDistinctNamespaces

brandond:
The current implementation will perform poorly when the desiredset
contains a large number of objects and namespaces, as it does a linear
scan of the string slice for each object. A more optimal approach is to
build a map of namespaces, and just return the map keys. This is much
faster since map operations are O(1).

Signed-off-by: Brad Davidson <brad.davidson@rancher.com>


pkg/apply/desiredset_process.go
```diff
@@ -466,22 +466,18 @@ func multiNamespaceList(ctx context.Context, namespaces []string, baseClient dyn
 	return wg.Wait()
 }
 
-func getDistinctNamespaces(objs map[objectset.ObjectKey]runtime.Object) (namespaces []string) {
+func getDistinctNamespaces(objs map[objectset.ObjectKey]runtime.Object) []string {
+	namespaces := map[string]bool{}
```
snasovich, line 470: Could use `map[string]struct{}` to save 1 byte per namespace. :)

pkg/apply/desiredset_process.go
```diff
@@ -466,22 +466,18 @@ func multiNamespaceList(ctx context.Context, namespaces []string, baseClient dyn
 	return wg.Wait()
 }
 
-func getDistinctNamespaces(objs map[objectset.ObjectKey]runtime.Object) (namespaces []string) {
+func getDistinctNamespaces(objs map[objectset.ObjectKey]runtime.Object) []string {
```
snasovich, line 469: Suggesting to apply the same change to `ObjectSet.Namespaces` - it's a bit of copy-paste there so may even extract helper func if you're up to it.

snasovich (changes requested):

snasovich (approved):
LGTM

pkg/apply/desiredset_process.go
brandond: I ended up tweaking things a bit more, wdyt?

snasovich: @brandond , I've noticed. ;) Like it. Just thought it would be great to add unit tests for `stringset.Set`.
Would like to have +1 approval anyway, will add couple more reviewers.

snasovich: Oh, didn't notice additional pushes (they arrived a few minutes after the comment). Will review later.

brandond: Yeah, tests for the new stringset are a good idea. Will add later.

kinarashah (approved):

brandond:
I think I'm set with this, if it looks good can someone merge?

PR #189, opened 2021-12-14
Ensure map keys are ordered before patch compare

thedadams:
In certain instances, marshaling a struct will produce a different result than marshaling a map with the same JSON object structure. This is because the encoding/json package guarantees ordered keys when marshaling a map. This would produce false change detection when calculating patches for certain objects.

After this change, the runtime.Object will be marshaled to bytes, unmarshaled to a map[string]interface{}, and then this map marshaled to bytes. This ensures that the keys will be ordered and will not produce unexpected patches.

Related issue:
https://github.com/rancher/rancher/issues/35819

### DIFF
--- a/pkg/apply/desiredset_compare.go
+++ b/pkg/apply/desiredset_compare.go
@@ -87,7 +87,7 @@ func originalAndModified(gvk schema.GroupVersionKind, oldMetadata v1.Object, new
 		return nil, nil, err
 	}
 
-	modified, err := json.Marshal(newObject)
+	modified, err := marshalWithOrderedKeys(newObject)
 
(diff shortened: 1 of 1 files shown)


thedadams:
For reviewers: this is from the Go documentation regarding order of keys when using `json.Marshal`:

> Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. **The map keys are sorted** and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:

thedadams:
This doesn't actually seem to fix the problem. I somehow go (un)lucky when testing this fix.

PR #190, opened 2021-12-15
Changing needacert to ignore service not found errors

MbolotSuse:
Webhooks/CRDS in needacert depend on the rancher-webhook service. The
OnService function ensures that they are updated when the service is
created. Therefore, when the service is not found, the err is no longer
returned, preventing undesired log messages and a retry loop.


MbolotSuse:
Related to [#33967](https://github.com/rancher/rancher/issues/33967)

pkg/needacert/needacert.go
```diff
@@ -112,7 +112,11 @@ func (h *handler) OnMutationWebhookChange(key string, webhook *adminregv1.Mutati
 		}
 
 		service, err := h.serviceCache.Get(webhookConfig.ClientConfig.Service.Namespace, webhookConfig.ClientConfig.Service.Name)
-		if err != nil {
+		if apierror.IsNotFound(err) {
+			//OnService will be called when the service is created, which will eventually update the webhook, so no
```
thedadams, line 116: nit: include a space between `//` and words, here an other places.

thedadams (approved):

pkg/needacert/needacert.go
MbolotSuse: Thanks for pointing this out. I've made the requested updates.

maxsokolovsky (approved):

pkg/needacert/needacert.go
```diff
@@ -218,7 +226,12 @@ func (h *handler) OnCRDChange(key string, crd *apiextv1.CustomResourceDefinition
 
 	service, err := h.serviceCache.Get(crd.Spec.Conversion.Webhook.ClientConfig.Service.Namespace,
 		crd.Spec.Conversion.Webhook.ClientConfig.Service.Name)
-	if err != nil {
+
+	if apierror.IsNotFound(err) {
+		// OnService will be called when the service is created, which will eventually update the CRD, so no
```
cbron, line 231: ~Shouldn't this be `OnCRDChange` ? OnService doesn't appear to call this.~ Looks like it does enqueue crd's

cbron (approved):

PR #194, opened 2022-01-20
Update kubernetes to v1.24

brandond:
Migrate to lease locks via the hybrid ConfigMapsLeasesResourceLock type.

This is extra fun because ConfigMapsResourceLock has been removed (or rather made private) in the same release that the replacement ConfigMapsLeasesResourceLock constant is introduced, so there's really no migration period available:
https://github.com/kubernetes/kubernetes/commit/29d9683cd0fb3cc81810a8b39715e5eb4c68b00e

This is required for Rancher to be used alongside Kubernetes 1.24 modules.


rosskirkpat (approved):

rosskirkpat:
Pasting some ideas from a slack discussion regarding this PR. This is a breaking change and we want to avoid the need to fork and maintain wrangler in two separate places.

Once this PR is merged, we should consider tagging wrangler as 1.0 and then split pre-1.24 development off to a release-v0.8 branch.

thedadams (approved):

brandond:
@thedadams could we get this merged today?

cmurphy (approved):

Issue #195, opened 2022-02-10
Wrangler v0.10.10 does not delete resources when applying an empty ObjectSet


brandond:
Application of nil DesiredSet does not delete existing resources. If I dry-run the apply and print the plan, it's clear that it's not finding the existing objects.

With wrangler 0.10.8:
```console
DEBU[0005] DeleteManifest plan: apply.Plan{
  Create:objectset.ObjectKeyByGVK{schema.GroupVersionKind{Group:"helm.cattle.io", Version:"v1", Kind:"HelmChart"}:[]objectset.ObjectKey(nil)}, 
  Delete:objectset.ObjectKeyByGVK{schema.GroupVersionKind{Group:"helm.cattle.io", Version:"v1", Kind:"HelmChart"}:[]objectset.ObjectKey{objectset.ObjectKey{Name:"traefik", Namespace:"kube-system"}, objectset.ObjectKey{Name:"traefik-crd", Namespace:"kube-system"}}}, 
  Update:apply.PatchByGVK{}, 
  Objects:[]runtime.Object(nil)
}
```

With wrangler 0.8.10:
```console
DEBU[0009] DeleteManifest plan: apply.Plan{
  Create:objectset.ObjectKeyByGVK{schema.GroupVersionKind{Group:"helm.cattle.io", Version:"v1", Kind:"HelmChart"}:[]objectset.ObjectKey(nil)}, 
  Delete:objectset.ObjectKeyByGVK{schema.GroupVersionKind{Group:"helm.cattle.io", Version:"v1", Kind:"HelmChart"}:[]objectset.ObjectKey{}}, 
  Update:apply.PatchByGVK{}, 
  Objects:[]runtime.Object(nil)
}
```

For some reason wrangler 0.8.10 isn't finding the existing objects to delete.

PR #196, opened 2022-02-10
Fix desiredset list with owner

brandond:
Related to
* https://github.com/rancher/wrangler/issues/195

When applying a desiredset with an owner, we need to check all namespaces for owned objects. Namespace pruning is only safe without ownership tracking.

### DIFF
--- a/pkg/apply/desiredset_process.go
+++ b/pkg/apply/desiredset_process.go
@@ -346,9 +346,11 @@ func (o *desiredSet) list(namespaced bool, informer cache.SharedIndexInformer, c
 	)
 
 	if informer == nil {
-		// if a lister namespace is set assume all objects belong to the listerNamespace
-		// otherwise use distinct namespaces from the desired objects
-		// (even if not namespaced as we'll get a single empty namespace in this case which is working OK)
+		// If a lister namespace is set, assume all objects belong to the listerNamespace.  If the
+		// desiredSet has an owner but no lister namespace, list objects from all namespaces to ensure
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

brandond:
Would appreciate if we could get a new tag cut once this is merged, so that we can pull it in to K3s and RKE2.

pkg/apply/desiredset_process.go
```diff
@@ -346,23 +346,38 @@ func (o *desiredSet) list(namespaced bool, informer cache.SharedIndexInformer, c
 	)
 
 	if informer == nil {
```
luthermonson, line 348: is this not a problem too when we have an informer and use the cache below?

brandond: The informer cache behavior wasn't changed as part of any of the previous PRs (https://github.com/rancher/wrangler/pull/183, https://github.com/rancher/wrangler/pull/187). It appears to only function with either non-namespaced objects, or when a namespace is provided. I'm assuming that matches its use since none of the controllers that use it with an informer were broken by the previous changes.

luthermonson: lgtm then

luthermonson (approved):

brandond:
@cmurphy I don't believe so; I tried to address that concern in the PR description:

> Setting an owner requires calling `apply.WithOwner(&ref).Apply(objs...)` so I don't think this affects the Rancher UI, as it does not provide a way for the user to supply an owner reference when applying raw YAML.

Applying from the UI was the original use case correct?

PR #203, opened 2022-04-22
Fix job reconciler diff

ibuildthecloud:
For reconcilers the originalObject that was passed is built from the the applied annotation.  That object will have all string values truncated to 64 bytes.  The job reconciler would fail because it does a full object compare and if any desired value was greater that 64 bytes it would never match because the originalObject's >64 byte value would be truncated.

To resolve this we now pass the newObject to be created and all the newObject truncated so that an exact comparsion can be done against the originalObject.

### DIFF
--- a/pkg/apply/reconcilers.go
+++ b/pkg/apply/reconcilers.go
@@ -24,7 +24,7 @@ var (
 	}
 )
 
-func reconcileDaemonSet(oldObj, newObj runtime.Object) (bool, error) {
+func reconcileDaemonSet(oldObj, _, newObj runtime.Object) (bool, error) {
 	oldSvc, ok := oldObj.(*appsv1.DaemonSet)
 	if !ok {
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
@luthermonson Can you or someone else take a look at this. Thanks.

luthermonson (approved):

pkg/apply/reconcilers.go
```diff
@@ -116,7 +116,7 @@ func reconcileService(oldObj, newObj runtime.Object) (bool, error) {
 	return false, nil
 }
 
-func reconcileJob(oldObj, newObj runtime.Object) (bool, error) {
+func reconcileJob(oldObj, newObj, _ runtime.Object) (bool, error) {
```
luthermonson, line 119: since youre only using the new prune obj here why not just round trip in this func and then we won't need to break the `Reconciler` api?

PR #236, opened 2022-11-08
[EXPERIMENTAL] Optimize Apply when WithIgnorePreviousApplied is specified

moio:
The `IgnorePreviousApplied` (aka `ignoreOriginal`) Apply option avoids the standard `kubectl apply`-inspired 3-way patching algorithm in favor of a simpler 2-way patch that ignores previous applications of the same object.

This patchset aims to avoid superfluous operations when this option is set while not introducing regressions.

This patchset is being tested by fleet users and will be finalized when there is conclusive evidence it actually helps with performance. I will remove the `[EXPERIMENTAL]` title at that point.

This patchset is best reviewed commit-by-commit.

### DIFF
--- a/pkg/apply/desiredset_compare.go
+++ b/pkg/apply/desiredset_compare.go
@@ -45,24 +45,26 @@ var (
 	}
 )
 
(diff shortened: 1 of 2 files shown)


moio:
Experiments did not reveal substantial benefits so far, closing.

PR #237, opened 2022-11-11
Update webhook matching

KevinJoiner:
The previous implementation of the match function used the RequestKind and RequestResource, which hold the GVK and GVR of the original object before any conversions were made. This negates the use of Kubernetes match policy.

If the creator of the handler function does not want to accept converted objects, they should update their webhook configuration and set the mathPolicy to `Exact`

Docs: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy


KevinJoiner:
Closing since this change is no longer needed by rancher/webhook

PR #240, opened 2022-11-15
leaderelection: exit 0 when exit requested

mattfarina:
The leader election code would always terminate fatally, with a non-zero exit code. Even if the application requested a shutdown. This was common when paired with wrangler-cli setting up a content that canceled when SIGTERM was detected. SIGTERM is sent when a node shuts down to tell the processes to exit, for example. The non-zero exit code is then recorded as an error by Kubernetes.

Instead, when an exit is requested because the context is set to done wrangler can cleanly exit. This is a pattern found in the kube scheduler

Relates to https://github.com/rancher/fleet/issues/1096 

SURE-5492

### DIFF
--- a/pkg/leader/leader.go
+++ b/pkg/leader/leader.go
@@ -58,7 +58,24 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
 				go cb(ctx)
 			},
 			OnStoppedLeading: func() {
-				logrus.Fatalf("leaderelection lost for %s", name)
+				select {
+				case <-ctx.Done():
+					// The context has been canceled or is otherwise complete.
+					// This is a request to terminate. Exit 0.
+					// Exiting cleanly is useful when the context is canceled
+					// so that Kubernetes doesn't record it exiting in error
+					// when the exit was requested. For example, the wrangler-cli
+					// package sets up a context that cancels when SIGTERM is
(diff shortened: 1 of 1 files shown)


mattfarina:
Note, I tested this with fleet and it works.

cmurphy (approved):

rmweir (approved):
The comment seems verbose and to be attempting to justify its own existence versus just explaining why it's there. Otherwise, LGTM

pkg/leader/leader.go
```diff
@@ -58,7 +58,24 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
 				go cb(ctx)
 			},
 			OnStoppedLeading: func() {
-				logrus.Fatalf("leaderelection lost for %s", name)
+				select {
+				case <-ctx.Done():
```
moio, line 62: Would this block if the context is not done, eg., when leader election was lost for any reason other than SIGTERM?

mattfarina: @moio It will not block. If the channel doesn't have something to read then that case isn't chosen. You can use select statements to look at multiple channels. You can read about it in a place more like https://hackthedeveloper.com/golang-select-multiple-channels/

moio: Great, I learnt something new today, thanks!

PR #257, opened 2023-01-12
Add codegen for EndpointSlices (discovery.k8s.io)

brandond:
EndpointSlices have been stable since 1.21, but we never got around to adding codegen for them because we're still using Endpoints everywhere. EndpointSlices are preferred for performance reasons; lets add codegen so that folks can start using them.

X-Ref: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/


jiaqiluo (approved):
LGTM 💯

kinarashah (approved):

snasovich (approved):

Issue #250, opened 2022-12-09
Transition to wrangler based release branches and more releases


mattfarina:
The current release branches are based on rancher major.minor versions. Since the 1.0.0 release of wrangler there are revisions of wrangler (not released versions) used in rancher.

Three problems have arisen from this approach:
1. Dependabot and other tools can't detect updates to apply other projects as updates. This means that those using wrangler (including the many rancher repos) need to do the updates manually.
2. The version of wrangler being used across the many consumers (many of those being other rancher codebases) are inconsistent in the version used. It's often not the latest. This means some bugs are not fixed uniformly across the codebase.
3. Releasing wrangler now has no value or meaning. And, it has not happened since the 1.0.0 release.

This proposal is to do the following:
- Use release branches in the form `release-vX.Y` where X is the major version and Y is the minor version. There are two reasons for this scheme.
  - The rancher release branches are in the form `release/vX.Y`. Doing the wrangler version release branches with the same scheme would eventually cause conflicts.
  - This scheme is already in use within this repo (i.e. `release-v0.8` already exists)
- All future releases would be cut from the release branch rather than master
- Once a release branch is created it's only for patch (i.e. bugfix releases). Those would first be merged into master and then cherry-picked into any release branches that should include it.
- The major/minor versions of wrangler to get continued work would be those that are in a supported release of rancher.

PR #263, opened 2023-01-30
Fix git reset command

KevinJoiner:
Porting git reset fix from release branches to master
https://github.com/rancher/wrangler/pull/261

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -309,7 +309,7 @@ func (g *Git) fetchAndReset(rev string) error {
 }
 
 func (g *Git) reset(rev string) error {
-	return g.git("-C", g.Directory, "reset", "--hard", "--", rev)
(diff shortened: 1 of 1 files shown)


pkg/git/git.go
```diff
@@ -309,7 +309,7 @@ func (g *Git) fetchAndReset(rev string) error {
 }
```
cbron, line 309: Change is fine, confused on branching though. If v1.0.2 was tagged off of branch release-v1.0, then what is master branch ? Maybe a head ? We could use more docs :)

macedogm (approved):

pkg/git/git.go
KevinJoiner: Matt has this issue for branching up https://github.com/rancher/wrangler/issues/250

KevinJoiner:
@cbron Can you merge this for me or add me as a maintainer?

PR #264, opened 2023-02-03
Generic controllers

KevinJoiner:
# Hack Week: Generic Controllers

Removed template code in `pkg/controller-gen/generators/type_go.go` that implements the controller and cache interfaces to call now call the generic implementation of Controllers and caches in `pkg/generic/controller.go` and `pkg/generic/cache.go` 

Added comments to our generated code to better increase rancher documentation.

Added a new package `pkg/generic/fake` which implements generic mocks for our controllers and clients.
You can check out the `pkg/generic/fake/README.md` for an example use of the mock.

# Notes on code changes
Most of the code has a 1:1 for one translation to generics with a few exceptions as shown below.

## Nil objects
In some of the logic, we need to check if the `runtime.Object` is nil with a simple `obj == nil`.
Generics can not be compared with a nil value since the value of T is not guaranteed to be a pointer. 
For those situations, you will see the following code, which compares the zero value of T to the object instead of nil. Since the zero value of a pointer is nil, we achieve the same desired results.
``` golang
var nilObj T
if obj == nilObj {}
```
## NonNamespaceControllers
Our generated code differs for resources that are namespace and ones that are not. For example this is the `Get(...)` function signature for a namespace and a non-namespaced resource.
```golang
Get(namespace, name string, options metav1.GetOptions) (*v1.Deployment, error)
```
``` golang
Get(name string, options metav1.GetOptions) (*v1.ClusterRole, error)
```
To provide the ability to satisfy both interfaces, there is a NoNamespaceController which inherits Controller and overrides
functions using a namespace, so they are called with an empty string
``` golang
func (c *NonNamespacedController[T, TList]) Get(name string, options metav1.GetOptions) (T, error) {
	return c.Controller.Get(metav1.NamespaceAll, name, options)
}
```

## Reflect.New(...)

When switching the client interface code to generics instead of generated code, I had to leverage the reflect package to create a new result object to return to the caller. Below we can see the two different implementations.
The use of reflect may bring up performance concerns, but after benchmarking, the two calls are the same speed.

generated
``` golang
func (c *deploymentController) Update(obj *v1.Deployment) (*v1.Deployment, error) {
	result := &v1.Deployment{}
	return result, c.client.Update(context.TODO(), obj.Namespace, obj, result, metav1.UpdateOptions{})
}
```
generic
``` golang
func (c *Controller[T, TList]) Update(obj T) (T, error) {
	result := reflect.New(c.objType).Interface().(T)
	return result, c.Client.Update(context.TODO(), obj.GetNamespace(), obj, result, metav1.UpdateOptions{})
}
```

```golang
package generic_test

import (
	"reflect"
	"testing"

	v1 "k8s.io/api/apps/v1"
)

var obj *v1.Deployment
var objType = reflect.TypeOf(obj).Elem()

func BenchmarkReflect(b *testing.B) {
	var result *v1.Deployment
	for n := 0; n < b.N; n++ {
		result = reflect.New(objType).Interface().(*v1.Deployment)
	}
	_ = result
}

func BenchmarkDeclared(b *testing.B) {
	var result *v1.Deployment
	for n := 0; n < b.N; n++ {
		result = &v1.Deployment{}
	}
	_ = result
}
```
```
go test -bench=. -benchtime=500000x 
goos: darwin
goarch: arm64
pkg: github.com/rancher/wrangler/pkg/generic
BenchmarkReflect-10               500000               305.4 ns/op
BenchmarkDeclared-10              500000               306.6 ns/op
PASS
ok      github.com/rancher/wrangler/pkg/generic 0.631s
```


pkg/generic/controller.go
```diff
@@ -22,3 +28,265 @@ type ControllerMeta interface {
...
+	Update(T) (T, error)
+
+	// UpdateStatus updates the Status field of a the object and return the newly updated Object or an error.
+	// Will always return an error if the object does not have a status field.
+	UpdateStatus(T) (T, error)
+
+	// Delete deletes the Object in the given name and Namespace.
+	Delete(namespace, name string, options *metav1.DeleteOptions) error
+
+	Get(namespace, name string, options metav1.GetOptions) (T, error)
```
cmurphy, line 106: missing doc comment

pkg/generic/controller.go
```diff
@@ -22,3 +28,265 @@ type ControllerMeta interface {
...
+	Patch(namespace, name string, pt types.PatchType, data []byte, subresources ...string) (result T, err error)
+}
+
+// ObjectHandler performs operations on the given runtime.Object and returns the new runtime.Object or an error
+type Handler func(key string, obj runtime.Object) (runtime.Object, error)
+
+// ObjectHandler performs operations on the given object and returns the new object or an error
+type ObjectHandler[T RuntimeMetaObject] func(string, T) (T, error)
+
+type Indexer[T RuntimeMetaObject] func(obj T) ([]string, error)
```
cmurphy: missing doc comment

cmurphy (approved):
This is really cool. I'm taking it a little on faith that any issues in this would be noticeable at compile time in Rancher, so lgtm

pkg/generic/controller.go
KevinJoiner: fixed

pkg/generic/controller.go
KevinJoiner: fixed

rmweir:
@KevinJoiner you mentioned a commit in your [commit message](https://github.com/rancher/wrangler/pull/264/commits/05d6b6dae23fbfde7c48f805645cca2794713637#). However, it doesn’t seem to exist and I think it’s too long anyway. Do you know what you intended to refer to here?

rmweir (changes requested):
I'm okay with the additional refactoring like removing the unused fields and the DeepCopyOnChange method. However, because these are exported this renders the PR a breaking change. We are now following proper semantic versioning for wrangler so this would put this change into a new major version. If you were to separate these changes I think there is a way for the switch to generics to go into the next minor.

pkg/generic/controller.go
```diff
@@ -22,3 +28,267 @@ type ControllerMeta interface {
...
+}
+
+// Watch will start watching resources in the given namespace.
+func (c *Controller[T, TList]) Watch(namespace string, opts metav1.ListOptions) (watch.Interface, error) {
+	return c.client.Watch(context.TODO(), namespace, opts)
+}
+
+// Patch will patch the resource with the matching name in the matching namespace.
+func (c *Controller[T, TList]) Patch(namespace, name string, pt types.PatchType, data []byte, subresources ...string) (T, error) {
+	result := reflect.New(c.objType).Interface().(T)
```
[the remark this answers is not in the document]
KevinJoiner: `var result T` returns the zero value for the `Object` interface, which is a nil pointer. In the PR message, I explain a bit more.

KevinJoiner:
@rmweir The function I removed was edited here 2044a7b2bb07a187dff7a936d6f6a8740d29ad9e and in the same PR #41 There is a commit dropping the function elsewhere 4a25b66cf1b74def1b341b446c36600ab723f9a9

pkg/generic/controller.go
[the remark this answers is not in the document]
rmweir: Got it. I appreciate the benchmarks!

rmweir (approved):

KevinJoiner:
Test Commit covers https://github.com/rancher/rancher/issues/39158

KevinJoiner:
Rebased For CI, If this looks good @rmweir can you merge for me.

PR #271, opened 2023-02-14
[DNM] Lasso options

KevinJoiner:
This PR is reliant on both PRs being merged first.  
https://github.com/rancher/wrangler/pull/264
https://github.com/rancher/lasso/pull/37

This will also be a breaking change, so we must figure out what to do with wrangler versioning.


go.mod
```diff
@@ -2,12 +2,16 @@ module github.com/rancher/wrangler
 
 go 1.19
 
+// Delete after lasso merge.
+replace github.com/rancher/lasso v0.0.0 => github.com/KevinJoiner/rancher-lasso v0.0.0-20230127201858-3076ace305aa
```
KevinJoiner, line 6: Marking this so we remember to remove it after the lasso PR is merged.

PR #272, opened 2023-02-15
Update readme

rmweir:
Related to:
https://github.com/rancher/wrangler/issues/250

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ Most people writing controllers are a bit lost as they find that there is nothin
 
 Wrangler is a framework for using controllers. Controllers wrap clients, informers, listers into a simple usable controller pattern that promotes some good practices.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,7 @@ Most people writing controllers are a bit lost as they find that there is nothin
 
 Wrangler is a framework for using controllers. Controllers wrap clients, informers, listers into a simple usable controller pattern that promotes some good practices.
 
+See info on branching strategy [here](https://github.com/rancher/wrangler/wiki/Wrangler-Branching-Strategy).
```
mattfarina, line 7: This is not what was previously agreed to which can be seen on https://github.com/rancher/wrangler/issues/250. I think we need further discussion.

mattfarina (changes requested):

PR #278, opened 2023-02-27
Updating go.mod for dirty repo issue

mattfarina:
Go commands are causing a dirty repo due to go.mod having things shifted around. A dirty repo stops CI. This change fixes the issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd
 	golang.org/x/sync v0.1.0
+	golang.org/x/text v0.7.0
 	golang.org/x/tools v0.6.0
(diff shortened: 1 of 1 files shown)


pkg/signals/signal_posix.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
rmweir, line 1: missing space between //
```suggestion
// go:build !windows
```

rmweir (changes requested):

pkg/signals/signal_posix.go
mattfarina: This new build tag IS NOT supposed to have a space. This comment was automatically generated by goimports based on the previous build tag. Some more details on the build tags at https://stackoverflow.com/a/68361186.

KevinJoiner (approved):

PR #279, opened 2023-02-28
Add documentation to name.Limit

manno:
* The func will crash with a runtime error for count <= 5.
* If count is 6, the result will have a leading dash.
* If len(string) == count, it will not return the original string, instead it appends a hash.


mattfarina:
@manno can you rebase. Testing works now. It was broken before.

mattfarina (approved):

Issue #249, opened 2022-12-09
Deprecate git package


mattfarina:
The git package within wrangler appears to only be used in two places and they are not using overlapping functionality. This package can be more easily maintained in those two separate places. So, this functionality should be documented as deprecated.

mattfarina:
Note, the git package can be removed in a 2.0.0 release of wrangler but not before (so we can follow semver)

KevinJoiner:
The git packaged has been removed from wrangler and the references have been removed from Rancher.

PR #289, opened 2023-03-28
Removes git pkg.

KevinJoiner:
Issue: https://github.com/rancher/wrangler/issues/249
Remove the git package for wrangler 2.0


rmweir:
Totally fine to go into master. More of a note that we have to be mindful that rancher properly replaces this before it updates to a version that contains this change. This will need to go in a new minor version or bundled with one that hasn't been tagged yet.

manno (approved):
Fleet already [removed the dependency](https://github.com/rancher/gitjob/pull/159) on the git package.

KevinJoiner:
@rmweir or @mattfarina, You should be good to merge this now since it is out of both Rancher and Fleet.

mattfarina (approved):

PR #299, opened 2023-05-05
readme: typo, if obj is nil, it should return directly

tjjh89017:
I'm not sure about this.
But I thought this is a typo.
If `obj` is `nil`, controller will not get/set any data to `obj`.


KevinJoiner (approved):

KevinJoiner:
@cbron Can you merge this? It is just a typo in the README

PR #303, opened 2023-06-12
Wrangler client with impersonation

KevinJoiner:
~~Dependent on https://github.com/rancher/lasso/pull/39~~

This first commit adds impersonation to the wrangler controllers. By allowing them to create a new client with impersonation.

The second 2 commits are updates to the generic controllers that make them easier to use. None of these changes should affect any code logic, only return types and interface definitions. 
These changes came about as I added generic controllers and mocks to the rancher project. I realized we could simplify our interfaces while still constraining to the expected interface definitions used by Rancher.

To add a bit of confidence you can see these changes used in rancher here https://github.com/rancher/rancher/pull/41817


go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/rancher/wrangler
 
 go 1.19
 
+// TODO: Update this dependency when is merged https://github.com/rancher/lasso/pull/39
+replace github.com/rancher/lasso v0.0.0-20221227210133-6ea88ca2fbcc => github.com/KevinJoiner/rancher-lasso v0.0.0-20230324131352-f9f8599f9af2
```
KevinJoiner, line 6: This will be removed when dependent PRs are added.

pkg/generic/controller.go
```diff
@@ -63,7 +63,7 @@ type ControllerInterface[T RuntimeMetaObject, TList runtime.Object] interface {
 	EnqueueAfter(namespace, name string, duration time.Duration)
 
 	// Cache returns a cache for the resource type T.
-	Cache() CacheInterface[T]
+	Cache() *Cache[T]
```
cmurphy, line 66: I'm sure you've covered this but could you explain just so I understand, if we have a test that uses a mocked ControllerInterface and it needs to call .Cache(), how does that returned cache get mocked if it's no longer an interface?

KevinJoiner: The short answer is we do not have any tests that do this. Mainly because the called to `Cache()` is usually called in a `Register(...)` or `New(...)` function which we currently don't test; you can see these updates to tests in rancher in this [PR](https://github.com/rancher/rancher/pull/41817)

My main reasoning for this change was to be more idiomatic, specifically after reading this section of [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments/#interfaces).
This change itself has a low impact on the functionality of this PR. So I am open to discussion if you still prefer returning the interface

pkg/controller-gen/generators/group_version_interface_go.go
```diff
@@ -93,7 +93,7 @@ func (f *groupInterfaceGo) Init(c *generator.Context, w io.Writer) error {
 		body := `
 		func (v *version) {{.type}}() {{.type}}Controller {
 			return &{{.type}}GenericController {
-				generic.New{{ if not .namespaced}}NonNamespaced{{end}}Controller[*{{.version}}.{{.type}}, *{{.version}}.{{.type}}List](schema.GroupVersionKind{Group: "{{.group}}", Version: "{{.version}}", Kind: "{{.type}}"}, "{{.pluralLower}}", {{ if .namespaced}}true, {{end}}v.controllerFactory),
+				generic.New{{ if not .namespaced}}NoNs{{end}}Controller[*{{.version}}.{{.type}}, *{{.version}}.{{.type}}List](schema.GroupVersionKind{Group: "{{.group}}", Version: "{{.version}}", Kind: "{{.type}}"}, "{{.pluralLower}}", {{ if .namespaced}}true, {{end}}v.controllerFactory),
```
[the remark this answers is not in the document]
KevinJoiner: Below are the differences between the two. I changed this after I had to type it out a couple of times. Honestly, both names are very verbose because you have to specify the types. So you are probably right in that we are not saving much space.
```golang
fake.NewMockNoNsControllerInterface[*v1.Namespace, *v1.NamespaceList](ctrl)
generic.NewNoNsController[*v1.Namespace, *v1.NamespaceList](schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"}, "namespaces", v.controllerFactory)

fake.NewMockNonNamespacedControllerInterface[*v1.Namespace, *v1.NamespaceList](ctrl)
generic.NewNonNamespacedController[*v1.Namespace, *v1.NamespaceList](schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"}, "namespaces", v.controllerFactory)
```

pkg/generic/controller.go
KevinJoiner: Actually after reading over the link I sent I realized that the rule applies to structs, not interfaces. So kept the change `generic.Controller` so it still returns a `generic.Cache` but reverted `generic.ControllerInterface` so that it returns `generic.CacheInterface`. This keeps us idiomatic while still allowing users to mock the `Cache(...)` function if needed.

KevinJoiner: Sorry for the whiplash, but the above can't be done because this statement does not compile 
```go
var ctrlInterface ControllerInterface[*v1.Pod, *v1.PodList] = &Controller[*v1.Pod, *v1.PodList]{}
```
```
have Cache() *generic.Cache[*"k8s.io/api/core/v1".Pod]
want Cache() generic.CacheInterface[*"k8s.io/api/core/v1".Pod]
```
So, I think we have to decided between
1. Return concrete types, and when we use them in our handler/controllers, we must have two member variables for Cache and Client separately for proper testing.
2.  Make an exception and continue returning an interface.

cmurphy: I think you're in a better position than I am to figure out the best option. What I'm worried about is a case where we might want to do something like `mockController.EXPECT().Cache().Return(...)`, or the equivalent in norman where the mock struct has a `ListerFunc`. Would there be a way to make those things work if `Cache()` returns a struct? In the golang document you linked, where would a case like this fit in those examples? I can see that there's nothing currently in rancher using either of those cases, but it kind of seems like a regression to say we could never test those cases if we wanted to.

KevinJoiner: > I can see that there's nothing currently in rancher using either of those cases, but it kind of seems like a regression to say we could never test those cases if we wanted to.

I agree with this. I don't think we gain much from being strict on the rule of not returning interfaces. Especially since it doesn't make the code easier to use. So I will convert it back to return an interface.

JonCrowther (approved):

pkg/generic/controller.go
```diff
@@ -46,7 +46,7 @@ type RuntimeMetaObject interface {
 }
 
 // ControllerInterface interface for managing K8s Objects.
-type ControllerInterface[T runtime.Object, TList runtime.Object] interface {
+type ControllerInterface[T RuntimeMetaObject, TList runtime.Object] interface {
```
cmurphy, line 49: These changes aren't from `go generate` are they? Should they be in a separate commit that explains the change?

KevinJoiner: They were needed so that the generated fake packages could be used with real Controllers and Caches.

KevinJoiner: I updated the commit message to better reflect the changes made

cmurphy (approved):

PR #305, opened 2023-06-19
leaderelection: configure all timeouts via environment variables

moio:
We have observed in `fleet-agent` user cases ([SURE-6125](https://jira.suse.com/browse/SURE-6125)/[fleet#1465](https://github.com/rancher/fleet/issues/1465)) that the leader election can be lost due to timeouts - not because pods are not working, but because the control plane is overwhelmed by load.

While of course having an overloaded control plane is a problem in and of itself, losing the election means that Wrangler-based pods will restart, which in turn means a new full listing of all interesting resources, creating additional load and exacerbating symptoms.

Ideally, the situation should improve when [API Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) goes out of beta and becomes commonplace - as it should ensure leader election API calls get served with higher priority, thus returning correctly within reasonable time frames even if the control plane is overloaded.

Meantime, pretty much the only other pragmatic option left is to give election more time before failure.

This PR allows to set all critical parameters via environment variables - with default values staying the same as before.

Note: this will also address ([SURE-3805](https://jira.suse.com/browse/SURE-3805)/[#1491](https://github.com/rancher/fleet/issues/1491))

### DIFF
--- a/pkg/leader/leader.go
+++ b/pkg/leader/leader.go
@@ -2,6 +2,7 @@ package leader
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
@@ -12,6 +13,13 @@ import (
 )
 
 type Callback func(cb context.Context)
+
+const defaultLeaseDuration = 45 * time.Second
+const defaultRenewDeadline = 30 * time.Second
+const defaultRetryPeriod = 2 * time.Second
+
+const developmentLeaseDuration = 45 * time.Hour
+const developmentRenewDeadline = 30 * time.Hour
 
 func RunOrDie(ctx context.Context, namespace, name string, client kubernetes.Interface, cb Callback) {
 	if namespace == "" {
@@ -43,16 +51,37 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
 		logrus.Fatalf("error creating leader lock for %s: %v", name, err)
 	}
 
-	t := time.Second
-	if dl := os.Getenv("CATTLE_DEV_MODE"); dl != "" {
-		t = time.Hour
+	leaseDuration := defaultLeaseDuration
+	renewDeadline := defaultRenewDeadline
+	retryPeriod := defaultRetryPeriod
+	if d := os.Getenv("CATTLE_DEV_MODE"); d != "" {
+		leaseDuration = developmentLeaseDuration
+		renewDeadline = developmentRenewDeadline
+	}
+	if d := os.Getenv("CATTLE_ELECTION_LEASE_DURATION"); d != "" {
+		leaseDuration, err = time.ParseDuration(d)
+		if err != nil {
+			return fmt.Errorf("CATTLE_ELECTION_LEASE_DURATION value [%s] is not a valid duration: %w", d, err)
+		}
+	}
+	if d := os.Getenv("CATTLE_ELECTION_RENEW_DEADLINE"); d != "" {
+		renewDeadline, err = time.ParseDuration(d)
+		if err != nil {
+			return fmt.Errorf("CATTLE_ELECTION_RENEW_DEADLINE value [%s] is not a valid duration: %w", d, err)
+		}
+	}
+	if d := os.Getenv("CATTLE_ELECTION_RETRY_PERIOD"); d != "" {
+		retryPeriod, err = time.ParseDuration(d)
+		if err != nil {
+			return fmt.Errorf("CATTLE_ELECTION_RETRY_PERIOD value [%s] is not a valid duration: %w", d, err)
+		}
 	}
 
 	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
 		Lock:          rl,
-		LeaseDuration: 45 * t,
-		RenewDeadline: 30 * t,
-		RetryPeriod:   2 * time.Second,
+		LeaseDuration: leaseDuration,
+		RenewDeadline: renewDeadline,
+		RetryPeriod:   retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(ctx context.Context) {
 				go cb(ctx)
--- a/pkg/leader/leader.go
+++ b/pkg/leader/leader.go
@@ -14,6 +14,11 @@ import (
 
 type Callback func(cb context.Context)
 
+const devModeEnvKey = "CATTLE_DEV_MODE"
+const leaseDurationEnvKey = "CATTLE_ELECTION_LEASE_DURATION"
+const renewDeadlineEnvKey = "CATTLE_ELECTION_RENEW_DEADLINE"
+const retryPeriodEnvKey = "CATTLE_ELECTION_RETRY_PERIOD"
+
 const defaultLeaseDuration = 45 * time.Second
 const defaultRenewDeadline = 30 * time.Second
 const defaultRetryPeriod = 2 * time.Second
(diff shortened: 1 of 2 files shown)


pkg/leader/leader.go
```diff
@@ -43,16 +51,37 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
...
+	if d := os.Getenv("CATTLE_ELECTION_RENEW_DEADLINE"); d != "" {
+		renewDeadline, err = time.ParseDuration(d)
+		if err != nil {
+			return errors.Wrapf(err, "CATTLE_ELECTION_RENEW_DEADLINE is not a valid duration")
+		}
+	}
+	if d := os.Getenv("CATTLE_ELECTION_RETRY_PERIOD"); d != "" {
+		retryPeriod, err = time.ParseDuration(d)
+		if err != nil {
+			return errors.Wrapf(err, "CATTLE_ELECTION_RETRY_PERIOD is not a valid duration")
```
rmweir, line 76: Nit: could include the value in these error messages
```suggestion
			return errors.Wrapf(err, "CATTLE_ELECTION_RETRY_PERIOD value [%s] is not a valid duration", d)
```

rmweir (changes requested):
One nit, otherwise LGTM.

pkg/leader/leader.go
moio: addressed as https://github.com/rancher/wrangler/commit/57d388e630d8c3d1726e7ef7e8bf1066eb12a930

moio:
@rmweir fixed. Do changes in wrangler require two approvers? If so, who could be the right person?

rmweir (approved):
LGTM

pkg/leader/leader.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/pkg/errors"
```
KevinJoiner, line 8: We do not want to use the package [github.com/pkg/errors](https://github.com/pkg/errors). It was made obsolete by go 1.13 when they added error wrapping to the standard lib. And the package has since been archived. Instead, we can use `fmt.Errorf("failed: %w", err)`

KevinJoiner (changes requested):

pkg/leader/leader.go
moio: Fixed

moio:
Well, the Bullseye Team does not have a _policy_ at this point.

There is a _guiding principle_, which is to adapt to each Team's patterns/habits/processes/policies to the extent reasonable and possible - the _goal_ being to help as good as possible.


Question becomes: would you normally add (or ask to add) tests in a case like this? If so, how would you propose to set them up? 

Currently code is basically a "main" method and the effect of setting environment variables is only time-detectable - not really amenable to unit tests. One possibility would be to refactor out the main added block:

https://github.com/rancher/wrangler/pull/305/files#diff-2b8e7c2af1963d0a67b49eb0af1b1107583a195b328f40db0fc28970d7ba4dc9R54-R77

to a separate function returning the three values or `err`, then test it the various error conditions. To my personal sensibility, that would not be worth the cost/benefit tradeoff, but you as the long-term maintainers are the ones to set the bar for contributions, so I will respect whatever decision you take here.

moio:
@KevinJoiner do you have a comment on the last update?

olblak:
@KevinJoiner , @rmweir What's the current state here? 
By looking at the discussion history, it seems that this issue is stalled for a month now

pkg/leader/leader.go
```diff
@@ -43,16 +51,37 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
...
+		}
+	}
+	if d := os.Getenv("CATTLE_ELECTION_RETRY_PERIOD"); d != "" {
+		retryPeriod, err = time.ParseDuration(d)
+		if err != nil {
+			return fmt.Errorf("CATTLE_ELECTION_RETRY_PERIOD value [%s] is not a valid duration: %w", d, err)
+		}
 	}
 
 	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
```
KevinJoiner, line 80: @olblak I was on paternity leave, so I was not watching updates.
@moio Our current guidelines are to add tests for all PRs, either Unit or Integration. Unit tests for this PR will have little immediate benefit but will help find regressions when someone goes to update this package in the future.
A good approach for adding tests would be to move most of this logic into a new function that returns a `leader.LeaderElectionConfig` and an `error` and then calls `leaderelection.RunOrDie(...)` with the output. This will allow future changes to the configuration to be easily tested as well.

We would get more mileage from creating Integration tests for this, but that is a much larger lift since there is no framework for it today.

KevinJoiner (changes requested):

moio:
@KevinJoiner: rebased, refactored per your suggestion, added unit tests. Please review again.

pkg/leader/leader_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	tests := []struct {
+		name    string
+		args    args
+		envs    []env
+		want    *leaderelection.LeaderElectionConfig
+		wantErr bool
+	}{
+		{
+			name: "all defaults",
```
KevinJoiner, line 30: This test and some of the others are dependent on running in a specific order since other tests may change the default env. If you create a resetEnv function that is run before every test and calls `os.Unsetenv` on all desired variables then we can have more deterministic results.

pkg/leader/leader.go
```diff
@@ -43,42 +51,75 @@ func run(ctx context.Context, namespace, name string, client kubernetes.Interfac
...
+	leaderelection.RunOrDie(ctx, *config)
 	panic("unreachable")
 }
+
+func computeConfig(rl resourcelock.Interface, cbs leaderelection.LeaderCallbacks) (*leaderelection.LeaderElectionConfig, error) {
+	leaseDuration := defaultLeaseDuration
+	renewDeadline := defaultRenewDeadline
+	retryPeriod := defaultRetryPeriod
+	var err error
+	if d := os.Getenv("CATTLE_DEV_MODE"); d != "" {
```
KevinJoiner, line 94: Nit: I know you didn't add this, but while we're here, can you move these string values to package level const.
```suggestion
	if d := os.Getenv(devModeKey); d != "" {
```

KevinJoiner (changes requested):

pkg/leader/leader_test.go
```diff
@@ -119,16 +119,23 @@ func Test_computeConfig(t *testing.T) {
...
+				{key: renewDeadlineEnvKey, value: "2s"},
+				{key: retryPeriodEnvKey, value: "bomb"},
 			},
 			want:    nil,
 			wantErr: true,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			for _, e := range os.Environ() {
```
KevinJoiner, line 132: First:
`os.Environ()` returns a list of envs in the form `key=value`. Instead I think you would want `os.Clearenv()`

Second
Unsetting all envs may be too aggressive, we should either explicitly unset the envs we care about Or capture the state of all the envs before our test and reset them after every test. 
Both of these options would avoid removing some standard shell envs like `PWD` that other unit tests/packages may be using.

KevinJoiner (changes requested):

pkg/leader/leader_test.go
moio: Right, addressed.

moio:
@rmweir do we already know when a new version of wrangler could be tagged? I am asking because of [a fleet issue](https://github.com/rancher/fleet/issues/1465) that would benefit from this change.

PR #309, opened 2023-06-29
CRD Generation updates

KevinJoiner:
## Issue https://github.com/rancher/rancher/issues/39335
These updates are to support the creation of CRDs from API structs instead of our custom CRD defs which we will no longer need to maintain.

### Commit 1
The first commit adds the ability to decode a YAML document into an object with JSON decoding. This is needed so we do not have to work with `unstructered.Unstuctered `Objects and map decoding.

### Commit 2 
Adds a new BatchCreate that creates CRDs from the apis structs. This is needed to support the new CRD generation in Rancher, which no longer use wrangler `crd.CRD`


KevinJoiner:
Closed in favor of https://github.com/rancher/wrangler/pull/310

PR #310, opened 2023-06-29
Crd generation

KevinJoiner:
Issue https://github.com/rancher/rancher/issues/39335
These updates are to support the creation of CRDs from API structs instead of our custom CRD defs which we will no longer need to maintain.

Commit 1
The first commit adds the ability to decode a YAML document into an object with JSON decoding. This is needed so we do not have to work with unstructered.Unstuctered Objects and map decoding.

Commit 2
Adds a new BatchCreate that creates CRDs from the apis structs. This is needed to support the new CRD generation in Rancher, which no longer use wrangler crd.CRD


pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create crd '%s': %w", crd.Name, err)
+		}
+		// item was not in initial list but does exist so we will attempt to get the latest resourceVersion
+		existingCRD, err = crdClient.Get(ctx, crd.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get exiting '%s' CRD: %w", crd.Name, err)
+		}
+		// fallthrough and attempt and update
```
JonCrowther, line 90:
```suggestion
		// fallthrough and attempt an update
```

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	// only keep the resource version for the desired object
+	crd.ResourceVersion = existingCRD.ResourceVersion
+	// if the CRD exist attempt to update it
```
JonCrowther, line 94:
```suggestion
	// if the CRD exists attempt to update it
```

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			return fmt.Errorf("failed waiting on CRD '%s': %w", crd.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// getExistingCRD returns a map of all CRD resource on the cluster.
+func getExistingCRD(ctx context.Context, crdClient clientv1.CustomResourceDefinitionInterface) (map[string]*apiextv1.CustomResourceDefinition, error) {
```
JonCrowther, line 58: Since this gets all existing CRDs, it should be plural
```suggestion
func getExistingCRDs(ctx context.Context, crdClient clientv1.CustomResourceDefinitionInterface) (map[string]*apiextv1.CustomResourceDefinition, error) {
```

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+	if !exist {
+		logrus.Infof("Creating embedded CRD %s", crd.Name)
+		_, err := crdClient.Create(ctx, crd, metav1.CreateOptions{})
+		if err == nil {
+			return nil
+		}
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create crd '%s': %w", crd.Name, err)
+		}
+		// item was not in initial list but does exist so we will attempt to get the latest resourceVersion
```
JonCrowther, line 85: In what situation would this happen? Where `crdClient.List` would not contain something that could be gotten via `crdClient.Get`?

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+	})
+}
+
+func crdIsReady(crd *apiextv1.CustomResourceDefinition) bool {
+	for i := range crd.Status.Conditions {
+		cond := &crd.Status.Conditions[i]
+		switch cond.Type {
+		case apiextv1.Established:
+			if cond.Status == apiextv1.ConditionTrue {
+				return true
```
JonCrowther, line 127: Just want to make sure this is intentional. If there are any `Established` status conditions that succeeded, we want to return true? No matter how many errors there may be?

pkg/crd/crd.go
KevinJoiner: 2 scenarios, both of which don't get hit right now.

1. Allow for a label selector in the List so we do not have to pull every CRD we may not care about.
I will update the BatchCreate call to take one in.

2. If the user does not have list access the logic can still be performed if you have the get verb.
This does not work right now because of your previous comment where I am returning the error.

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// ensure each CRD exist
+	for _, crd := range crds {
+		err = ensureCRD(ctx, crd, crdClient, existingCRDs)
+		if err != nil {
+			return err
+		}
+	}
+
+	existingCRDs, err = getExistingCRD(ctx, crdClient)
```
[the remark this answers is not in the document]
KevinJoiner: 1. I forgot to stop returning an error here.
2. The second list is to get the updated versions of the CRDs. After the creation or update of the CRDs K8s will add a status to the resource, letting us know if it is ready or not. We make the second list so we can get that status information and determine if we need to wait for the CRD to be ready with `waitCRD(...)` or not.

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
+// Package crd helps with the installation of CustomResourceDefinitions.
```
maxsokolovsky, line 1: Tiny nit: I'd add a more descriptive or concrete comment, not sure what "helps" really means.

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+
+// getExistingCRD returns a map of all CRD resource on the cluster.
+func getExistingCRD(ctx context.Context, crdClient clientv1.CustomResourceDefinitionInterface) (map[string]*apiextv1.CustomResourceDefinition, error) {
+	storedCRDs, err := crdClient.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list crds: %w", err)
+	}
+
+	// convert existingCRDs to a map for faster lookup
+	existingCRDs := map[string]*apiextv1.CustomResourceDefinition{}
```
maxsokolovsky, line 65: You could pre-allocate the capacity of the map.

pkg/crd/crd.go
KevinJoiner: Yes, the `condition.Type` is a unique key for all of the conditions in the list.
You may be curious why we don't return false if the condition is not true. This is because we also want to log if the CRD has a naming conflict for the case below, so we have to loop through the entire list to check for both.

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get exiting '%s' CRD: %w", crd.Name, err)
+		}
+		// fallthrough and attempt and update
+	}
+	// only keep the resource version for the desired object
+	crd.ResourceVersion = existingCRD.ResourceVersion
+	// if the CRD exist attempt to update it
+	logrus.Infof("Updating embedded CRD %s", crd.Name)
+	_, err := crdClient.Update(ctx, crd, metav1.UpdateOptions{})
```
maxsokolovsky, line 96: Since you update the CRD in the cluster, would you want to also update it in your in-memory map?

KevinJoiner: This map is not reused after this, so there is no need to update it.

pkg/crd/crd.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+
+	existingCRDs, err = getExistingCRD(ctx, crdClient)
+	if err != nil {
+		return fmt.Errorf("unable to check for existing CRDs: %w", err)
+	}
+	// ensure each CRD is ready
+	for _, crd := range crds {
+		if existingCRD, ok := existingCRDs[crd.Name]; ok && crdIsReady(existingCRD) {
```
maxsokolovsky, line 44: I think you already know that the CRD exists in the map, since you'd run `ensure` above?

KevinJoiner: Correct I added a label selector to the list so there is a chance the CRD does not have that label and is not returned in the list.

pkg/yaml/yaml.go
```diff
@@ -24,15 +26,62 @@ var (
 	cleanContains = []string{
 		"cattle.io/",
 	}
+	buffSize = 4096
```
maxsokolovsky, line 29: This may be a constant.

pkg/yaml/yaml.go
```diff
@@ -24,15 +26,62 @@ var (
...
 )
 
+// Unmarshal decodes YAML bytes into document (as defined by
+// the YAML spec) then converting it to JSON via
+// "k8s.io/apimachinery/pkg/util/yaml".YAMLToJSON and then decoding the json in the the v interface{}
 func Unmarshal(data []byte, v interface{}) error {
 	return yamlDecoder.NewYAMLToJSONDecoder(bytes.NewBuffer(data)).Decode(v)
 }
 
+// UnmarshalWithJSONDecoder takes a reader of yaml bytes converts the document or documents to JSON
```
maxsokolovsky, line 39: Could you tighten up the comment a bit? I am being so annoying because it's an exported function. 😺

```suggestion
// UnmarshalWithJSONDecoder expects a reader of raw YAML. 
// It converts the document or documents to JSON,
// then decodes the JSON bytes into a slice of values of type T.
// Type T must be a pointer, or the function will panic.
```

pkg/yaml/yaml.go
```diff
@@ -24,15 +26,62 @@ var (
 	cleanContains = []string{
 		"cattle.io/",
 	}
+	buffSize = 4096
 )
 
+// Unmarshal decodes YAML bytes into document (as defined by
+// the YAML spec) then converting it to JSON via
+// "k8s.io/apimachinery/pkg/util/yaml".YAMLToJSON and then decoding the json in the the v interface{}
 func Unmarshal(data []byte, v interface{}) error {
```
maxsokolovsky, line 35: Is this function similar to the new one? We just shouldn't use it anymore because it's less safe the new, generic one?

pkg/yaml/yaml_test.go
```diff
@@ -0,0 +1,161 @@
+package yaml
```
maxsokolovsky, line 1: You could make all tests and subtests run in parallel.

maxsokolovsky (changes requested):

pkg/crd/crd_test.go
```diff
@@ -0,0 +1,308 @@
+// Package crd helps with the installation of CustomResourceDefinitions.
```
maxsokolovsky, line 1: Could you please update this package comment as well? This one seems to win when it comes to our godoc - there are two package comments now.

pkg/crd/crd.go
```diff
@@ -0,0 +1,143 @@
...
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	clientv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var (
+	waitDuration = 1 * time.Minute
```
maxsokolovsky, line 19: Feel free to ignore this nit - but I would make these constants here and pass them to your waitForCRD function. They seem to be variables just so they can be changed in tests, which is OK, but not ideal.

KevinJoiner: Since `time.Minute` is technically a struct the compiler won't allow me to make it constant.

pkg/crd/crd_test.go
KevinJoiner: missed this one.

pkg/crd/crd.go
maxsokolovsky: It will, since values of type `time.Duration` are really just numbers.

pkg/yaml/yaml.go
KevinJoiner: Can do it's a good suggestion 👍🏾

pkg/crd/mockCRDClient_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	gomock "github.com/golang/mock/gomock"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	v10 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	types "k8s.io/apimachinery/pkg/types"
+	watch "k8s.io/apimachinery/pkg/watch"
+)
+
+// MockCustomResourceDefinitionInterface is a mock of CustomResourceDefinitionInterface interface.
+type MockCustomResourceDefinitionInterface struct {
```
maxsokolovsky, line 19: A general question: why use mocks here? Couldn't you create a fake implementation instead?

pkg/crd/crd_test.go
```diff
@@ -0,0 +1,308 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+var (
+	errTest = errors.New("test Error")
+)
+
+func TestBatchCreateCRDs(t *testing.T) {
```
maxsokolovsky, line 24: This test and its subtests could also be run in parallel.

pkg/yaml/yaml.go
KevinJoiner: The main difference is the new one works with YAML, which contains a list of docs or a single doc. `Unmarshal(...)` only works for bytes that contain a single document. The other function that already exists to help with lists is `ToObject(...)`, but this function forces you to use an `unstructured.Unstructured` objects which as you know cause a lot of headaches when you try to convert them to actual objects with map decode.

`Unmarshal(...)` is still safe if you know your YAML file is not a list of objects. `UnmarshalWithJSONDecoder(...)` provides more functionality, I don't want to depreciate the function since if you are using it today there is no reason to change.

pkg/crd/crd.go
KevinJoiner: Oh nice, I didn't think it would.  I am going to leave it since it is not exported, and it keeps the test from taking a minute to run when testing the public `BatchCreate(..)` function.

maxsokolovsky: I was thinking you could pass it as a parameter, rather than use the package variable, but it's up to you.

pkg/crd/init.go
```diff
@@ -361,6 +363,35 @@ func (c CRD) ToCustomResourceDefinitionV1Beta1() (*apiextv1beta1.CustomResourceD
 	return v1beta1CRD, nil
 }
 
+// Name resolves the Name for the give CRD.
+func (c *CRD) Name() string {
```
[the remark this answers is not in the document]
KevinJoiner: I just updated the method to do less.

pkg/crd/crd.go
KevinJoiner: Oh, you mean to the `BatchCreateCRDs()` sure I am fine with that.

maxsokolovsky: Or rather `waitCRD` could take it. And `BatchCreateCRDs` would pass it...

pkg/crd/mockCRDClient_test.go
KevinJoiner: The fake.SimpleClient does not allow me to track what methods were called and with what parameters, the same way mocks do. My first implementation used the fake mocks from K8s but I ended up having to add custom tracking code and the test were pretty hard to follow.

pkg/crd/init.go
```diff
@@ -361,6 +363,31 @@ func (c CRD) ToCustomResourceDefinitionV1Beta1() (*apiextv1beta1.CustomResourceD
 	return v1beta1CRD, nil
 }
 
+// Name resolves the Name for the give CRD.
```
JonCrowther, line 366:
```suggestion
// Name resolves the Name for the given CRD.
```

JonCrowther (approved):

maxsokolovsky (approved):

PR #311, opened 2023-06-30
Merge release branch

KevinJoiner:
currently, the latest release tag is not in the master branch because the release branch was never merged back upstream. This means when you add a master commit to a dependent go.mod it will get reverted to point to the latest release by go mod tidy, which is the older release commit 8a1858d3883c7e0eed81204b614397472449627f


KevinJoiner:
We need to do the merge this way because a direct merge won't pass CI due to the Fossa changes see https://github.com/rancher/wrangler/pull/312

PR #317, opened 2023-09-01
Update v2

rmweir:


KevinJoiner (approved):

KevinJoiner (approved):

pkg/codegen/main.go
```diff
@@ -19,6 +19,7 @@ import (
 
 func main() {
 	controllergen.Run(args.Options{
+		ImportPackage: "github.com/rancher/wrangler/v2/pkg/generated",
 		OutputPackage: "github.com/rancher/wrangler/pkg/generated",
```
mattfarina, line 23: Does the `OutputPackage` also need the `v2` added?

KevinJoiner: `OutputPackage` determines where in the repo to put the generated files. I had the same question before and tested it out locally, and when you change the output to `v2` it creates a new local v2 directory for the generated file. I think it would be more clear if the variable name was `OutputPath` instead of package.

mattfarina: That makes sense.

mattfarina (approved):

KevinJoiner (approved):

PR #321, opened 2023-09-18
Add support for getting array index in GetValue

cmurphy:
Add the ability to use an index number to get an item out of a slice.

The number will still be a string representation of an integer. For
example, to get the last element from a nested slice like:

```
data := map[string]interface{}{
    "data": []interface{}{
        "first",
        "second",
        "third",
    },
}
```

Use keys like

```
val := GetValues(data, "dat", "2")
```

https://github.com/rancher/rancher/issues/42767

### DIFF
--- a/pkg/data/values.go
+++ b/pkg/data/values.go
@@ -1,5 +1,9 @@
 package data
 
+import (
+	"strconv"
+)
+
 func RemoveValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
 	for i, key := range keys {
 		if i == len(keys)-1 {
@@ -18,16 +22,36 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
 	return val
 }
 
-func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
+func GetValue(data interface{}, keys ...string) (interface{}, bool) {
 	for i, key := range keys {
 		if i == len(keys)-1 {
-			val, ok := data[key]
-			return val, ok
+			if dataMap, ok := data.(map[string]interface{}); ok {
+				val, ok := dataMap[key]
+				return val, ok
+			}
+			if dataSlice, ok := data.([]interface{}); ok {
+				return itemByIndex(dataSlice, key)
+			}
 		}
-		data, _ = data[key].(map[string]interface{})
+		if dataMap, ok := data.(map[string]interface{}); ok {
+			data, _ = dataMap[key]
+		} else if dataSlice, ok := data.([]interface{}); ok {
+			data, _ = itemByIndex(dataSlice, key)
+		}
 	}
 
 	return nil, false
+}
+
+func itemByIndex(dataSlice []interface{}, key string) (interface{}, bool) {
+	keyInt, err := strconv.Atoi(key)
+	if err != nil {
+		return nil, false
+	}
+	if keyInt >= len(dataSlice) || keyInt < 0 {
+		return nil, false
+	}
+	return dataSlice[keyInt], true
 }
 
 func PutValue(data map[string]interface{}, val interface{}, keys ...string) {
--- a/pkg/data/values.go
+++ b/pkg/data/values.go
@@ -1,9 +1,15 @@
+// Package data contains functions for working with unstructured values like []interface or map[string]interface{}.
+// It allows reading/writing to these values without having to convert to structured items.
 package data
 
 import (
 	"strconv"
 )
 
+// RemoveValue removes a value from data. Keys should be in order denoting the path to the value in the nested
+// structure of the map. For example, passing []string{"metadata", "annotations"} will make the function remove the
+// "annotations" key from the "metadata" sub-map. Returns the removed value (if any) and a bool indicating if the value
+// was found.
 func RemoveValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
 	for i, key := range keys {
 		if i == len(keys)-1 {
@@ -22,6 +28,11 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
 	return val
 }
 
+// GetValue retrieves a value from the provided collection, which must be a map[string]interface or a []interface.
+// Keys are always strings.
+// For a map, a key denotes the key in the map whose value we want to retrieve.
+// For the slice, it denotes the index (starting at 0) of the value we want to retrieve.
+// Returns the retrieved value (if any) and a bool indicating if the value was found.
 func GetValue(data interface{}, keys ...string) (interface{}, bool) {
 	for i, key := range keys {
 		if i == len(keys)-1 {
@@ -54,6 +65,8 @@ func itemByIndex(dataSlice []interface{}, key string) (interface{}, bool) {
 	return dataSlice[keyInt], true
 }
 
+// PutValue updates the value of a given map at the index specified by keys that denote the path to the value in the
+// nested structure of the map. If there is no current entry at a key, a new map is created for that value.
 func PutValue(data map[string]interface{}, val interface{}, keys ...string) {
 	if data == nil {
 		return
--- a/pkg/data/values.go
+++ b/pkg/data/values.go
@@ -28,12 +28,25 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
 	return val
 }
 
-// GetValue retrieves a value from the provided collection, which must be a map[string]interface or a []interface.
+// GetValue works similar to GetValueFromAny, but can only process maps. Kept this way to avoid breaking changes with
+// the previous interface, GetValueFromAny should be used in most cases since that can handle slices as well.
+func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
+	for i, key := range keys {
+		if i == len(keys)-1 {
+			val, ok := data[key]
+			return val, ok
+		}
+		data, _ = data[key].(map[string]interface{})
+	}
+	return nil, false
+}
+
+// GetValueFromAny retrieves a value from the provided collection, which must be a map[string]interface or a []interface.
 // Keys are always strings.
 // For a map, a key denotes the key in the map whose value we want to retrieve.
 // For the slice, it denotes the index (starting at 0) of the value we want to retrieve.
 // Returns the retrieved value (if any) and a bool indicating if the value was found.
-func GetValue(data interface{}, keys ...string) (interface{}, bool) {
+func GetValueFromAny(data interface{}, keys ...string) (interface{}, bool) {
 	for i, key := range keys {
 		if i == len(keys)-1 {
 			if dataMap, ok := data.(map[string]interface{}); ok {
--- a/pkg/data/values_test.go
+++ b/pkg/data/values_test.go
@@ -1,0 +1,94 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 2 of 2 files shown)


pkg/data/values_test.go
```diff
@@ -0,0 +1,108 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetValue(t *testing.T) {
```
maxsokolovsky, line 9: Please run the test and all its subtests in parallel.

pkg/data/values_test.go
```diff
@@ -0,0 +1,108 @@
...
+	tests := []struct {
```
maxsokolovsky, line 10: Nit: you may want to add a test case where the top-level `data` input is a slice, not a map.

JonCrowther (approved):
LGTM, one question for my personal understanding

maxsokolovsky (changes requested):

pkg/data/values_test.go
cmurphy: Passing in a slice as the top level is not supported, this would not compile unless we changed the function signature.

pkg/data/values_test.go
```diff
@@ -0,0 +1,108 @@
...
+			name: "key is in first level of map",
+			data: map[string]interface{}{
+				"realKey": "realVal",
+			},
+			keys:        []string{"realKey"},
+			wantValue:   "realVal",
+			wantSuccess: true,
+		},
+		{
+			name: "key is nested in map",
```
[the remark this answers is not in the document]
cmurphy: No, there is no valid path for parent -> grandhild -> child, it will stop processing when it doesn't find "grandchild" under "parent". Adding a test to show this.

pkg/data/values_test.go
cmurphy: done

pkg/data/values_test.go
```diff
@@ -0,0 +1,123 @@
...
+					},
+				},
+			},
+			keys:        []string{"parent", "children", "-3"},
+			wantValue:   nil,
+			wantSuccess: false,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
maxsokolovsky, line 116: You'd need to shadow the test variable in the first statement inside the for loop.
`test := test`

pkg/data/values_test.go
maxsokolovsky: You did already change the function signature, such that it accepts `interface{}`.

cmurphy: Indeed I did and somehow forgot, adding the test

pkg/data/values_test.go
```diff
@@ -0,0 +1,135 @@
...
+						"eve",
+					},
+				},
+			},
+			keys:        []string{"parent", "children", "2"},
+			wantValue:   "eve",
+			wantSuccess: true,
+		},
+		{
+			name: "get index of top levelslice",
```
maxsokolovsky, line 85:
```suggestion
			name: "get index of top level slice",
```

What about a case where there is a slice of maps with each map having a key whose value is a slice of elements? I think it might be good to add...

pkg/data/values.go
```diff
@@ -18,18 +22,38 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
 	return val
 }
 
-func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
+func GetValue(data interface{}, keys ...string) (interface{}, bool) {
```
KevinJoiner, line 25: I don't love this addition of optionally taking a slice and a number key as a string. Can we just make a new function for slices?

KevinJoiner (changes requested):

pkg/data/values.go
cmurphy: I could revert this function to only take a map and to call out to another function when it encounters a nested slice, that covers the use case I care about. But I can't completely separate the map and slice functions or have it not take a number as a string, because the data is unstructured and the caller (steve) doesn't know whether it is going to encounter a slice.

MbolotSuse: Resolving, decision made offline to allow a slice and a string index value.

pkg/data/values_test.go
MbolotSuse: I've adjusted the name of this test, added the requested test case, and added another test case which attempted to use a string as a slice index value. Please re-open if this doesn't meet your suggestion here.

pkg/data/values.go
```diff
@@ -18,18 +27,43 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
 	return val
 }
 
-func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
+// GetValue gets a value from the provided data field, which must be a map[string]interface or a []interface. Keys can
```
maxsokolovsky, line 30: I would tweak the doc comment like so:

```
// GetValue retrieves a value from the provided collection], which must be a map[string]interface or []interface. 
// Keys are always strings. 
// For the map, they denote the path to the value within the nested structure of the map.
// For the slice, they denote the index.
// Returns the retrieved value (if any) and a bool indicating if they value was found.
```

pkg/data/values_test.go
```diff
@@ -0,0 +1,170 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetValue(t *testing.T) {
```
maxsokolovsky, line 9: I think we might want to have a few more test cases.

Where the function receives: 
1. a slice and multiple keys denoting indices. This will help to be clearer on the doc comment for the func, as I don't know what the behavior we get now.
1. a slice and an empty string as index
1. a map and an empty string as index
1. a slice and no second argument
1. a map and no second argument

pkg/data/values.go
```diff
@@ -1,5 +1,14 @@
+// Package data contains functions for working with unstructured values like interfaces or map[string]interface{}
```
maxsokolovsky, line 1:
```suggestion
// Package data contains functions for working with unstructured values like []interface and map[string]interface{}.
// It allows reading and mutating them without having to convert them to structured values.
```

pkg/data/values.go
```diff
@@ -18,18 +27,43 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
...
+	if err != nil {
+		return nil, false
+	}
+	if keyInt >= len(dataSlice) || keyInt < 0 {
+		return nil, false
+	}
+	return dataSlice[keyInt], true
+}
+
+// PutValue puts val into data (in place) at the index specified by keys (specified in order of precedence). If there
```
maxsokolovsky, line 65:
```suggestion
// PutValue updates the value of a given map at the index specified by keys that denote the path to the value in the nested structure of the map.
// If there is no current entry at a key, a new map is created for that value.
```

pkg/data/values.go
```diff
@@ -1,5 +1,14 @@
+// Package data contains functions for working with unstructured values like interfaces or map[string]interface{}
+// allows reading/writing to these values without having to convert to structured items.
 package data
 
+import (
+	"strconv"
+)
+
+// RemoveValue removes a value from data. Keys should be in order of precedence (i.e. "metadata", "annotations") means
```
maxsokolovsky, line 9:
```suggestion
// RemoveValue removes a value from data. Keys should be in order denoting the path to the value in the nested structure of the map.
// For example, passing []string{"metadata", "annotations"} will make the function remove the "annotations" key from the "metadata" sub-map.
```

maxsokolovsky (changes requested):
A few doc suggestions and additional test cases.

pkg/data/values_test.go
MbolotSuse: 1. This is covered by [this test](https://github.com/rancher/wrangler/blob/c29289d99068cbd11024b470165423927dcc881c/pkg/data/values_test.go#L96). If you are looking for something different, let me know.

2-5 - I'll take a look at it, those seem reasonable.

MbolotSuse: I've added the tests for 2-5. Let me know if you see anything else here.

pkg/data/values.go
MbolotSuse: I've made some adjustments to the doc comment. LMK if you want any further changes.

pkg/data/values.go
MbolotSuse: I've made some adjustments to the doc comment. LMK if you want any further changes.

pkg/data/values.go
MbolotSuse: I've made some adjustments to the doc comment. LMK if you want any further changes.

pkg/data/values.go
MbolotSuse: I've made some adjustments to the doc comment. LMK if you want any further changes.

pkg/data/values_test.go
maxsokolovsky: I think the test you link is slightly different as it works with slices and maps, where some keys are for the map. But a case where there is just a slice and multiple keys denoting indices is different.
The collection would be just a flat slice and the keys are indices. I am curious as to which index is the one chosen for the return value.

maxsokolovsky (dismissed):

JonCrowther (approved):

MbolotSuse (changes requested):
@tomleb pointed out that although the function is technically backward compatible, the signature no longer is. To make this backward compatible and not force a release of wrangler v3.0, we will need to keep the interface compatible. I'll do this by making the current `GetValue` in this PR a new function, and keeping the original function consistent.

pkg/data/values.go
```diff
@@ -18,18 +28,51 @@ func GetValueN(data map[string]interface{}, keys ...string) interface{} {
...
 func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
+	return GetValueFromAny(data, keys...)
+}
+
+// GetValueFromAny retrieves a value from the provided collection, which must be a map[string]interface or a []interface.
+// Keys are always strings.
+// For a map, a key denotes the key in the map whose value we want to retrieve.
+// For the slice, it denotes the index (starting at 0) of the value we want to retrieve.
+// Returns the retrieved value (if any) and a bool indicating if the value was found.
+func GetValueFromAny(data interface{}, keys ...string) (interface{}, bool) {
```
tomleb, line 42: Wonder whether it would be a good idea to allows  `keys` to be `any` and validate that they're `string` or a number..
This way we could be sure that `-1` passed in expected a list vs `"-1"` which would expect a map.

This does add more complexity with this function like what to return if you receive an unexpected type, etc. So I think it's okay as is.

tomleb (approved):

pkg/data/values.go
MbolotSuse: I think that the way this function is used in Steve is that keys are essentially user provided input. So if we wen that direction Steve would need to do some selective parsing there, which I don't think is desirable. 

That being said, I think that you are correct that the underlying library really should be more specific than this. I tried to implement a more comprehensive approach that (IIRC) did what you suggested in #325 (though I doubt that will merge). I'd say if we want an improvement like that we should consider a rewrite like that Pr looks for.

tomleb (approved):
LGTM. I think we could add tests for the `GetValue` function now that it's not relying on `GetValueFromAny`, but up to you.

MbolotSuse (approved):

JonCrowther (approved):

PR #325, opened 2023-10-12
Adding field logic to data package

MbolotSuse:
Added a new Field struct which provides the ability to search through unstructed data (such as the data returned from a json call), and easily add or remove values to this data.

Related to #321.


brandond:
I'm curious how these differ from what you can do with [`metav1.Unstructured`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) 

For your example of `RemoveValue(example, "hello", "nested")` - could this not also be done with `RemoveNestedField(example, "hello", "nested") from the unstructured package?

MbolotSuse:
@brandond The big answer would be slice support - from the docs from that page (for example [SetNestedField](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#SetNestedField)):

>  SetNestedField sets the value of a nested field to a deep copy of the value provided. Returns an error if value cannot be set because one of the nesting levels is not a map[string]interface{}.

From testing as well, this appears to return an error if you try and recurse through a slice like:
```go
test := map[string]any{
	"test": []any{"hello"},
}
err := unstructured.SetNestedField(test, "hello2", "test", "0")
fmt.Printf("%+v, err: %v \n", test, err) //prints an error
```

There's also [SetNestedSlice](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#SetNestedSlice), but that appears to be for setting a slice value, not for navigating through a slice. 

In short, this can process through slices, that library you linked doesn't seem to be able to do that.

PR #326, opened 2023-10-13
Use indexer when listing cached types by hash

aruiz14:
Relates to rancher/fleet#1842

Extending `WithCacheType` option, those types will be indexed by their hash label (`objectset.rio.cattle.io/hash`), which every object produced or modified by `Apply` has. This is a direct replacement for using `List` with an exact label selector, which is certainly more expensive.
The ultimate goal of this change is reducing CPU usage, which has been successfully observed by testing this change in a custom Fleet build (along with rancher/fleet#1857), where in some cases produced around 40%-50% CPU drop (see SURE-6990).

### DIFF
--- a/pkg/apply/desiredset.go
+++ b/pkg/apply/desiredset.go
@@ -2,16 +2,23 @@ package apply
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	"github.com/rancher/wrangler/v2/pkg/apply/injectors"
 	"github.com/rancher/wrangler/v2/pkg/kv"
 	"github.com/rancher/wrangler/v2/pkg/merr"
 	"github.com/rancher/wrangler/v2/pkg/objectset"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/cache"
 )
+
+// Indexer name added for cached types
+const byHash = "wrangler.byObjectSetHash"
 
 type patchKey struct {
 	schema.GroupVersionKind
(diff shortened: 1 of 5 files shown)


pkg/apply/desiredset.go
```diff
@@ -160,13 +168,47 @@ func (o desiredSet) WithCacheTypes(igs ...InformerGetter) Apply {
...
+			}
+		}
+		pruneTypes[ig.GroupVersionKind()] = informer
 	}
 
 	o.pruneTypes = pruneTypes
 	return o
 }
 
+// AddHashIndexer adds an indexer to each provided informer that indexes objects by the hash annotation value
```
moio, line 185: Please expand this description with the reasons why a developer should add this option and ideally a usage example.

aruiz14: Updated, there was no reason to export this function, thanks!

rmweir (changes requested):
Small nit then LGTM.

pkg/apply/desiredset.go
```diff
@@ -2,17 +2,25 @@ package apply
...
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	"github.com/rancher/wrangler/v2/pkg/apply/injectors"
 	"github.com/rancher/wrangler/v2/pkg/kv"
 	"github.com/rancher/wrangler/v2/pkg/merr"
 	"github.com/rancher/wrangler/v2/pkg/objectset"
+
```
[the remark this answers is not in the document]
aruiz14: It's not an arbitrary newline. Imports are grouped by stdlib/local/external. Nonetheless, I've change it to stick to the current convention, thanks.
c3044b9

pkg/apply/desiredset_process.go
```diff
@@ -494,3 +498,45 @@ func multiNamespaceList(ctx context.Context, namespaces []string, baseClient dyn
 
 	return wg.Wait()
 }
+
+// canIndexByHash detects if provided selector can be replaced by using the hash index, if configured, in which case returns the hash value
+func canIndexByHash(indexer cache.Indexer, selector labels.Selector) (string, bool) {
```
KevinJoiner, line 503: nit:
You can ignore this because it is opinionated and not a big deal, but this function caused some confusion for me when I was first reading the code.
prefixes such as `can`, `is`, and `has` usually imply that we are returning a bool, while this function is more of a retrieval function that lets you know if the value was found or not.
```suggestion
func getIndexByHash(indexer cache.Indexer, selector labels.Selector) (string, bool) {
```

pkg/apply/desiredset_process_test.go
```diff
@@ -109,3 +113,157 @@ func Test_multiNamespaceList(t *testing.T) {
...
+			indexer: cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{byHash: func(obj interface{}) ([]string, error) {
+				return nil, nil
+			}}),
+			selector: selector,
+		}, wantHash: hash, want: true},
+		{name: "non exact hash selector", args: args{
+			indexer: cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{byHash: func(obj interface{}) ([]string, error) {
+				return nil, nil
+			}}),
+			selector: labels.Everything(),
```
KevinJoiner, line 148: nit:
labels.Everything() does not have any Requirements so it fails early before checking if the selector has the desired hash key.

pkg/apply/desiredset_process_test.go
```diff
@@ -109,3 +113,157 @@ func Test_multiNamespaceList(t *testing.T) {
 		})
 	}
 }
+
+func Test_canIndexByHash(t *testing.T) {
```
KevinJoiner, line 117: There is nothing to change here; it's just a suggestion for future tests. We would get a lot of mileage by having one test function for the entire `list(...)` function since we can better verify the desired interaction with the current behavior.

KevinJoiner (approved):
Two minor nits, but they don't change the validity of the changes here.

pkg/apply/desiredset_process_test.go
aruiz14: Nice catch, thanks!

aruiz14:
@KevinJoiner I had to upgrade the `golangci-lint` GH action because it was failing for every commit (I even tested re-running it for an old one that already succeeded in the past).
Please let me know if I should revert it and maybe create a separate PR.

KevinJoiner (approved):
You can either squash all the commits other than the lint update and force push or we can just squash everything when we merge.

aruiz14:
> You can either squash all the commits other than the lint update and force push or we can just squash everything when we merge.

Done! Thanks for your review!

rmweir (approved):
LGTM

moio:
@KevinJoiner I see I cannot merge this, can you help?

moio (approved):

PR #327, opened 2023-10-20
WIP: FastApply

moio:
This PR implements a faster Apply option that uses 2-way merge instead of 3-way merge.

```golang
// WithFastApply configures Apply to use 2-way merging, which is less CPU intensive than the default 3-way merging.
// WithFastApply has the same behavior as Apply (including to honor strategic merge patch tags) apart from one case:
//
// When applying an object with a nil field, WithFastApply will clear the corresponding field in the cluster object.
// If that is not wanted, specify the field in dotted notation in dontClearFields, and it will be left unchanged.
//
// Instead, when applying an object with a nil field, basic Apply will decide whether to clear it or not depending on
// the **immediately previous** Apply. If the field set nil was also nil in the immediately preceeding Apply, then
// it is not cleared, otherwise it is.
func (a *apply) WithFastApply(dontClearFields ...string) Apply {
```

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -82,6 +82,7 @@ type Apply interface {
 	WithSetOwnerReference(controller, block bool) Apply
 	WithIgnorePreviousApplied() Apply
 	WithDiffPatch(gvk schema.GroupVersionKind, namespace, name string, patch []byte) Apply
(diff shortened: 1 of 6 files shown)


moio:
Approach has been abandoned for the time being, might be revisited later. Until that point I am closing this PR.

PR #341, opened 2023-11-23
Allow GeneratingHandlers to skip Apply if resource version didn't change

aruiz14:
Extends `GeneratingHandlerOptions` to allow a new `UniqueApplyForResourceVersion`, which will keep track of the latest resource version for each individual resource that was handled, avoiding subsequent calls to Apply if the first one was successful. The goal is to limit the number of Apply is called with the same inputs, reducing the computation costs. However, this means that handlers must be deterministic, in the sense that they must produce the same object list for the same input resource, hence this option must be carefully enabled on a case by case basis. It has no impact on other users if the option is not enabled.

We have successfully tested this feature on Fleet's Bundle and GitRepo generating handlers (see rancher/fleet#1964 for an example usage), while trying to mitigate the effect of over-firing handlers (we are also working towards the root cause of the over-firing, which may be partially caused by the use of `Enqueue`).

### DIFF
--- a/pkg/controller-gen/generators/type_go.go
+++ b/pkg/controller-gen/generators/type_go.go
@@ -98,10 +98,14 @@ type {{.type}}Cache interface {
 }
 
 {{ if .hasStatus -}}
+// {{.type}}StatusHandler is executed for every added or modified {{.type}}. Should return the new status to be updated
 type {{.type}}StatusHandler func(obj *{{.version}}.{{.type}}, status {{.version}}.{{.statusType}}) ({{.version}}.{{.statusType}}, error)
 
+// {{.type}}GeneratingHandler is the top-level handler that is executed for every {{.type}} event. It extends {{.type}}StatusHandler by a returning a slice of child objects to be passed to apply.Apply
 type {{.type}}GeneratingHandler func(obj *{{.version}}.{{.type}}, status {{.version}}.{{.statusType}}) ([]runtime.Object, {{.version}}.{{.statusType}}, error)
 
+// Register{{.type}}StatusHandler configures a {{.type}}Controller to execute a {{.type}}StatusHandler for every events observed.
+// If a non-empty condition is provided, it will be updated in the status conditions for every handler execution
 func Register{{.type}}StatusHandler(ctx context.Context, controller {{.type}}Controller, condition condition.Cond, name string, handler {{.type}}StatusHandler) {
 	statusHandler := &{{.lowerName}}StatusHandler{
 		client:    controller,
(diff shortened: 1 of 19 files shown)


rmweir (approved):
LGTM

pkg/controller-gen/generators/type_go.go
```diff
@@ -200,14 +205,38 @@ func (a *{{.lowerName}}GeneratingHandler) Handle(obj *{{.version}}.{{.type}}, st
...
 
-	return newStatus, generic.ConfigureApplyForObject(a.apply, obj, &a.opts).
+	err = generic.ConfigureApplyForObject(a.apply, obj, &a.opts).
 		WithOwner(obj).
 		WithSetID(a.name).
 		ApplyObjects(objs...)
+	if err == nil {
+		a.seenResourceVersion(obj)
+	}
+	return newStatus, err
```
KevinJoiner, line 219: Because this code is written in a template and not actually go code, We should keep logic simple even if it costs an extra line.
```suggestion
	if err != nil {
		return newStatus, err
	}
	a.seenResourceVersion(obj)
	return newStatus, nil
```

pkg/controller-gen/generators/type_go.go
```diff
@@ -200,14 +205,38 @@ func (a *{{.lowerName}}GeneratingHandler) Handle(obj *{{.version}}.{{.type}}, st
 	}
 
 	objs, newStatus, err := a.{{.type}}GeneratingHandler(obj, status)
-	if err != nil {
+	if err != nil || !a.isNewResourceVersion(obj) {
 		return newStatus, err
 	}
```
KevinJoiner, line 210: Similar to my other comment, this is logically the same it better describes our intent since we are maintaining templated code.
```suggestion
	if err != nil {
		return newStatus, err
	}
	if !a.isNewResourceVersion(obj) {
	    return newStatus, nil
	}
```

pkg/controller-gen/generators/type_go.go
```diff
@@ -200,14 +205,38 @@ func (a *{{.lowerName}}GeneratingHandler) Handle(obj *{{.version}}.{{.type}}, st
...
+		return true
+	}
+
+	// Apply once per resource version
+	key := obj.Namespace + "/" + obj.Name
+	previous, ok := a.seen.Load(key)
+	return !ok || previous != obj.ResourceVersion
+}
+
+func (a *{{.lowerName}}GeneratingHandler) seenResourceVersion(obj *{{.version}}.{{.type}}) {
```
KevinJoiner, line 233: nit:
```suggestion
func (a *{{.lowerName}}GeneratingHandler) storeResourceVersion(obj *{{.version}}.{{.type}}) {
```

pkg/generic/generating_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/rancher/wrangler/v2/pkg/apply"
+	fakeapply "github.com/rancher/wrangler/v2/pkg/apply/fake"
+	v1 "github.com/rancher/wrangler/v2/pkg/generated/controllers/core/v1"
+	fake2 "github.com/rancher/wrangler/v2/pkg/generic/fake"
+)
+
+func TestUniqueApplyForResourceVersion(t *testing.T) {
+	const n = 3
```
KevinJoiner, line 20: nit: make this variable more descriptive
```suggestion
	const numOfHandlerCalls = 3
```

KevinJoiner (changes requested):
Overall, this looks good. I don't love adding more code in generated templates, but this is a good approach to solving the problem.

Also, I didn't mark this but I would be forever grateful if you added function comments to GeneratingHandler since you are already in the package.

aruiz14:
> Also, I didn't mark this but I would be forever grateful if you added function comments to GeneratingHandler since you are already in the package.

I'm sorry I missed this comment when I first addressed your feedback. It's done now, please take another look, thanks!

KevinJoiner (approved):

PR #351, opened 2024-01-03
[v1.1] Allow GeneratingHandlers to skip Apply if resource version didn't change

aruiz14:
Backport of #341. Note that this pull request was originally forward-ported from v1, as that is what Fleet 0.7 and Fleet 0.8 uses. This makes this patch the closest to what was already tested by some users on their testing and production setups (using a debug build). In other words, this backport (and subsequent new v1 release) is necessary in order to use this change on Fleet, since the current release branches still use wrangler/v1, and upgrading to v2 is not feasible.

Besides updating the import paths, the only change from the original PR is in the test code:
Since the `pkg/generic/fake` does not exist in `v1` (used in tests), I had to replace it by using a mock


aruiz14:
I needed this change in `v1` in order to be able to use it in Fleet (`master` branch already upgraded to `v2`, but existing release branches still use `v1`).

aruiz14:
CI is consistently failing but it's unrelated to my changes. I've opened #352 to address it

moio (approved):
Functional changes seem identical to the original PR - I will let others comment on test code if necessary. LGTM

KevinJoiner (approved):

mattfarina (approved):

PR #378, opened 2024-04-05
Support multiple base branches with renovate

JonCrowther:
Continuation of https://github.com/rancher/rancher/issues/44968

Renovate supports the ability to target multiple base branches. It will update the dependencies of those branches separately, so we don't have to backport any dependency fixes. 

More information on this renovate feature can be found here:
https://docs.renovatebot.com/configuration-options/#basebranches

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,7 +3,9 @@
     "github>rancher/renovate-config#release"
   ],
   "baseBranches": [
-    "master"
+    "master",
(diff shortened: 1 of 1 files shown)


maxsokolovsky (approved):

tomleb (approved):
LGTM, let's test it out to see what happens. Before merging any of these though we'll see what kind of policies we want to have wrt. older versions.

PR #382, opened 2024-04-09
Dependencies bumps

tomleb:
Dependencies bump for these PRs:
- #376 
- #375 
- #374


JonCrowther (approved):

tomleb:
@MbolotSuse Is this something we could merge in master, considering it would go in 2.2.0-rc3, which would end up in 2.8?

Otherwise, we'd need to create a release-2.2 branch and then this could end up in master...

MbolotSuse:
@tomleb Let's merge this in master. I don't see an issue with this potentially going in 2.2.0 given all of the CVE bumps that will already go in that release. It doesn't need to if we don't end up creating any new RCs for 2.2.0, but I don't have an overall issue with it.

PR #416, opened 2024-04-18
Use goimports instead of k8s.io/code-generator

galal-hussein:
There is a conflict happens in kubernetes 1.30:

```
go: github.com/k3s-io/k3s/pkg/codegen imports
	github.com/rancher/wrangler/pkg/controller-gen/args imports
	k8s.io/code-generator/cmd/client-gen/path: module k8s.io/code-generator@latest found (v0.30.0, replaced by github.com/k3s-io/kubernetes/staging/src/k8s.io/code-generator@v1.30.0-k3s1), but does not contain package k8s.io/code-generator/cmd/client-gen/path
```

because upstream has removed the path package in 1.30, this fix will mitigate the use of said package and should outputs the same result.

### DIFF
--- a/pkg/controller-gen/args/groupversion.go
+++ b/pkg/controller-gen/args/groupversion.go
@@ -7,9 +7,9 @@ import (
 	"strings"
 
 	"golang.org/x/tools/go/packages"
+	"golang.org/x/tools/imports"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/code-generator/cmd/client-gen/generators/util"
-	"k8s.io/code-generator/cmd/client-gen/path"
 	"k8s.io/gengo/types"
 )
 
@@ -70,7 +70,7 @@ func toVersionType(obj interface{}) (string, *types.Name) {
 	if t.Kind() == reflect.Ptr {
 		t = t.Elem()
 	}
-	pkg := path.Vendorless(t.PkgPath())
+	pkg := imports.VendorlessPath(t.PkgPath())
 	return versionFromPackage(pkg), &types.Name{
 		Package: pkg,
 		Name:    t.Name(),
@@ -118,7 +118,7 @@ func ScanDirectory(pkgPath string) (result []Type, err error) {
 		for i := 0; i < s.NumFields(); i++ {
 			f := s.Field(i)
 			if f.Embedded() && f.Type().String() == "k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta" {
-				pkgPath := path.Vendorless(pkgs[0].PkgPath)
+				pkgPath := imports.VendorlessPath(pkgs[0].PkgPath)
 				result = append(result, Type{
 					Package: pkgPath,
 					Version: versionFromPackage(pkgPath),


brandond (approved):

dereknola (approved):

ericpromislow (changes requested):
This doesn't work for me (ubuntu, golang 1.22.1):
```
$ git diff
index 76490e7..c9e7358 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/rancher/wrangler/v2
 
-go 1.20
+go 1.22.0
+
+toolchain go1.22.1
 
 require (
  github.com/evanphx/json-patch v5.6.0+incompatible
@@ -16,9 +18,9 @@ require (
  golang.org/x/tools v0.20.0
  k8s.io/api v0.28.6
  k8s.io/apiextensions-apiserver v0.28.6
- k8s.io/apimachinery v0.28.6
+ k8s.io/apimachinery v0.30.0
  k8s.io/client-go v0.28.6
- k8s.io/code-generator v0.28.6
+ k8s.io/code-generator v0.30.0
  k8s.io/gengo v0.0.0-20220902162205-c0856e24416d
...
iff --git a/pkg/controller-gen/args/groupversion.go b/pkg/controller-gen/args/>
index 40cc78b..1297713 100644
--- a/pkg/controller-gen/args/groupversion.go
+++ b/pkg/controller-gen/args/groupversion.go
@@ -7,9 +7,9 @@ import (
  "strings"
 
  "golang.org/x/tools/go/packages"
+ "golang.org/x/tools/imports"
  "k8s.io/apimachinery/pkg/runtime/schema"
  "k8s.io/code-generator/cmd/client-gen/generators/util"
- "k8s.io/code-generator/cmd/client-gen/path"
  "k8s.io/gengo/types"
 )
 
@@ -70,7 +70,7 @@ func toVersionType(obj interface{}) (string, *types.Name) {
  if t.Kind() == reflect.Ptr {
    t = t.Elem()
  }
- pkg := path.Vendorless(t.PkgPath())
+ pkg := imports.VendorlessPath(t.PkgPath())
  return versionFromPackage(pkg), &types.Name{
    Package: pkg,
    Name:    t.Name(),
@@ -118,7 +118,7 @@ func ScanDirectory(pkgPath string) (result []Type, err erro>
    for i := 0; i < s.NumFields(); i++ {
      f := s.Field(i)
      if f.Embedded() && f.Type().String() == "k8s.io/apimachinery/pkg/apis/met>
-       pkgPath := path.Vendorless(pkgs[0].PkgPath)
+       pkgPath := imports.VendorlessPath(pkgs[0].PkgPath)
        result = append(result, Type{
          Package: pkgPath,
          Version: versionFromPackage(pkgPath),
$ make build
go build ./...
# github.com/rancher/wrangler/v2/pkg/controller-gen
pkg/controller-gen/main.go:57:3: cannot use clientgenerators.NameSystems(nil) (value of type "k8s.io/gengo/v2/namer".NameSystems) as "k8s.io/gengo/namer".NameSystems value in argument to genericArgs.Execute
pkg/controller-gen/main.go:183:32: undefined: dpargs.CustomArgs
pkg/controller-gen/main.go:209:32: undefined: csargs.NewDefaults
pkg/controller-gen/main.go:248:52: undefined: cs.Packages
pkg/controller-gen/main.go:311:33: undefined: infargs.NewDefaults
pkg/controller-gen/main.go:327:53: undefined: inf.Packages
pkg/controller-gen/main.go:335:20: undefined: lsargs.NewDefaults
pkg/controller-gen/main.go:349:52: undefined: ls.Packages
make: *** [Makefile:11: build] Error 1
$ ls  -l ~/go/pkg/mod/k8s.io/code-generator@v0.*/cmd/deepcopy-gen/args
/home/ericp/go/pkg/mod/k8s.io/code-generator@v0.25.4/cmd/deepcopy-gen/args:
total 4
-r--r--r-- 1 ericp ericp 1815 Feb 16 10:15 args.go

/home/ericp/go/pkg/mod/k8s.io/code-generator@v0.27.9/cmd/deepcopy-gen/args:
total 4
-r--r--r-- 1 ericp ericp 1815 Jan 16 16:07 args.go

/home/ericp/go/pkg/mod/k8s.io/code-generator@v0.29.3/cmd/deepcopy-gen/args:
total 4
-r--r--r-- 1 ericp ericp 1815 Apr  2 17:10 args.go

/home/ericp/go/pkg/mod/k8s.io/code-generator@v0.30.0/cmd/deepcopy-gen/args:
total 4
-r--r--r-- 1 ericp ericp 1619 Apr 23 13:03 args.go
$ diff3 ~/go/pkg/mod/k8s.io/code-generator@v0.2*/cmd/deepcopy-gen/args/args.go
<no output>
```

galal-hussein:
@ericpromislow I think this is an issue with wrangler and its compatibility with 1.30 as well, there are few commits happened with 1.30 that restructured the controller-gen and its cmds, for example https://github.com/k3s-io/kubernetes/commit/b40e4d7e7059a989c9b940ab13bc4d77e199116f, this will require other changes to wrangler to adapt to 1.30

ericpromislow (approved):
Actually it compiles fine if I don't bump the package to 0.30.0, so approving it, realizing that there will be more work when we do bump it.

brandond:
So just to be clear we will drop the deprecated lib here, and then in another separate PR we will add Kubernetes v1.30 support. That SECOND PR will be a breaking change for Wrangler, and will require forking again to maintain branches of Wrangler on either side of the v1.30 codegen split.

tomleb:
A concern here is with wrangler's semantic versioning. The two functions are _slight different_ and could result in different code gen (afaik). See https://go.dev/play/p/hjoh5YyX-9y. If we consider this breaking then we'd have to bump wrangler.

> That SECOND PR will be a breaking change for Wrangler, and will require branching again to maintain releases of Wrangler on either side of the v1.30 codegen split.

There might be a way to avoid a breaking change in wrangler. Due to semantic versioning that wrangler follows, it would require us to bump wrangler to v3, which is not ideal. (As an aside, we're working on an ADR on wrangler (and other repos) versioning strategy.)

If we can keep the same API within `pkg/controller-gen`, `pkg/controller-gen/args` and `pkg/controller-gen/generators` then that wouldn't be breaking change (assuming the codegen output is also the same).

MbolotSuse (approved):
LGTM, though this is technically a breaking change, so it will require a new lasso major version.

brandond:
I don't see that a minor change in codegen to drop a deprecated library in favor of a drop-in replacement would be a breaking change that would require a new major version. There have been changes to codegen in the past that have been made across patch or minor releases of this library.

The additional work for v1.30 to use gengo/v2 will definitely require a major version bump.

galal-hussein:
I agree with @brandond, I think the upcoming changes will require a major version bump, however this change is really not a breaking change in my opinion

tomleb:
Closing, included in https://github.com/rancher/wrangler/pull/425.

PR #425, opened 2024-04-30
K8s 1.30

tomleb:
# Issue https://github.com/rancher/rancher/issues/45324

Wrangler relies on `k8s.io/code-generator` for _some_ of its generators. Starting with k8s 1.30, these generators have been migrated,  in a backward incompatible way, from `k8s.io/gengo` to `k8s.io/gengo/v2`. We must adapt the code with the new major version of gengo.

# Changes in `gengo/v2`

## GeneratorArgs

One major change in the API is that before, you had to create a [GeneratorArgs](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/args/args.go#L51). It contained information such as: [InputDirs](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/args/args.go#L53) to point to directories that need codegen, [OutputBase](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/args/args.go#L56) to specify a directory tree to write file to, [CustomArgs](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/args/args.go#L85) for generator-specific information.

This struct is no longer used in `gengo/v2`, and instead is replaced in the following way:
- `InputDirs`: This is now provided in a new argument to the [Execute function](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/v2/execute.go#L76).

- `OutputBase`: Generators would compute "`output file = <output base>/<package>/...`" to know which directory the generated file needs to be in. By convention, generators now expects an absolute path to the destination directory. For example:
```diff
-	args.OutputPackagePath = filepath.Join(customArgs.Package, "clientset")
+	clientSetArgs.OutputDir = filepath.Join(customArgs.OutputBase, customArgs.Package, "clientset")
```
- `CustomArgs`: Custom args aren't part of the API anymore. If the `Target`/`Generator` needs custom arguments, you must find another way of passing it to it. By convention it's done by having a `GetTarget` function that accepts such an argument, and wrapping it by a function that doesn't accept it. Here's an example:
```diff
-	if err := genericArgs.Execute(
-		clientgenerators.NameSystems(nil),
-		clientgenerators.DefaultNameSystem(),
-		clientGen.Packages,
+	getTargets := func(context *generator.Context) []generator.Target {
+		return clientGen.Packages(context, customArgs)
+	}
```
## DefaultGen

[generator.DefaultGen](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/generator/default_generator.go#L33) is replaced by  [generator.GoGenerator](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/v2/generator/go_generator.go#L33). The impact is that we must specify `.go` suffix to all filenames ourselves. This changes the generator name to have this `.go` suffix, which itself changes some comments in the generated output.

## DefaultPackage

[generator.DefaultPackage](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/generator/default_package.go#L24) is replaced by [generator.SimpleTarget](https://github.com/kubernetes/gengo/blob/51d4e06bde7070d095fd4c266e51dbe135511a97/v2/generator/simple_target.go#L25).

# Commits

To make it easier to review, I have split the changes in multiple commits. Not every commit is self-contained (so don't try to `go build` them individually).

**1. go-imports**

Cherry-picked from https://github.com/rancher/wrangler/pull/416.

**2. Upgrade k8s to v1.30**

Bumps the following:
- k8s.io to v1.30
-  [gengo to the same commit](https://github.com/kubernetes/kubernetes/blob/release-1.30/go.mod#L231) used by this kubernetes release.
- Go to 1.22 (same Go as kubernetes 1.30)
- Lasso to a version that supports kubernetes 1.30

**3. DeepCopy generator**

Not much changed here. We're now referencing gengo directly instead of [k8s.io/code-generator's newtype](https://github.com/kubernetes/code-generator/blob/release-1.29/cmd/deepcopy-gen/args/args.go#L28) We were already using the generator from `gengo/v1`, which is still available to us.

**4. ClientSet generator**

Migrates the clientset generator to gengo/v2.

**5. Listers generator**

Migrates the listers generator to gengo/v2.

**6. Informers generator**

Migrates the informers generator to gengo/v2.

**7. setGenClient middleware**

I haven't change the logic of this function. It acts as some kind of middleware. It used to passthrough the arguments to the next function. Since that's no longer part of the `gengo/v2` api, we're no longer doing this.

**8. Migrate custom generators**

As mentioned in the "Changes in `gengo/v2`" section, generators now embed a `generator.GoGenerator` instead. Anything referencing `Package` should now be named `Target`. For example, I renamed the function `Package` to `Target` and it now returns a `SimpleTarget`. 

The boilerplate information is read only once as opposed to be read for every generator. The goal wasn't to optimize, but to avoid passing the `GeneratorArgs` from `gengo/v1`.


pkg/controller-gen/args/args.go
```diff
@@ -12,6 +12,8 @@ type CustomArgs struct {
 	TypesByGroup  map[schema.GroupVersion][]*types.Name
 	Options       Options
 	OutputBase    string
+	// TODO: Rename?
+	Boilerplate []byte
```
tomleb, line 16: I'm gonna try to remove this to make it as it was before if possible.

MbolotSuse (changes requested):
There's one other place that we should update the go version of: https://github.com/rancher/wrangler/blob/2a768fdce92ff55c3f08639708dc11c9cdfa3e4a/.github/workflows/lint.yaml#L4. Would be good to give that a look as well.

pkg/controller-gen/args/args.go
tomleb: As discussed previously, I ended up renaming this to `BoilerplateContent` instead and added a comment.

Ideally, `CustomArgs` would be private and wouldn't embed `Options`, but that's not the case yet :shrug:

tomleb:
Recent changes:
- Rebased from latest wrangler master
- Pulled in latest lasso with k8s 1.30 support
- Renamed `Boilerplate` to `BoilerplateContent`
- Updated the Go version in the linting script

MbolotSuse (commented):
Two comments on how we are structuring the package path.

galal-hussein (approved):
LGTM

pkg/controller-gen/main.go
```diff
@@ -206,11 +206,11 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
...
 	}
 
-	args, clientSetArgs := csargs.NewDefaults()
+	clientSetArgs := csargs.New()
 	clientSetArgs.ClientsetName = "versioned"
-	args.OutputBase = customArgs.OutputBase
-	args.OutputPackagePath = filepath.Join(customArgs.Package, "clientset")
-	args.GoHeaderFilePath = customArgs.Options.Boilerplate
+	clientSetArgs.OutputDir = filepath.Join(customArgs.OutputBase, customArgs.Package, "clientset")
+	clientSetArgs.OutputPkg = filepath.Join(customArgs.Package, "clientset")
```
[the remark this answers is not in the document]
MbolotSuse: Resolving, the imports seem to be backward compatible since we move the generated after they have been made.

MbolotSuse (approved):

PR #426, opened 2024-04-30
Migrate wrangler to GHA

ericpromislow:
Related to: [#45226](https://github.com/rancher/rancher/issues/45226) - for the master branch


tomleb:
Do we want arm64 support in the CI? Any reason for removing it?

ericpromislow:
What's the `arm64` equivalent of running on `ubuntu-latest`? Or better still, how would I specify both arm64 and amd64 runs?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
+name: Wrangler CI
+
+on:
+  push:
+  pull_request:
+    tags:
+      - v*
+    branches:
+      - 'release/*'
```
tomleb, line 9: The branches are named differently for wrangler. The currently supported branches are:
- `master`
- `release-2.2`
- `release-2.1`
- `release-2.0-untagged`

```suggestion
      - 'release-*'
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  ci:
+    strategy:
+      matrix:
+        arch:
+        - amd64
+        - arm64
+    runs-on: org-${{ github.repository_owner_id }}-${{ matrix.arch }}-k8s
+    container: registry.suse.com/bci/golang:1.22
```
tomleb, line 20: This container already comes with golang built-in, but we're also installing Go in a later step. How do we know which Go version is being used? Maybe this container should just be a regular ubuntu image (similar to ubuntu-latest, but for self-hosted runners). I kinda prefer to keep the `actions/setup-go` action just because we don't need to update the Go version, but I'm not against removing the step either.

ericpromislow: Good point - no need to install go

.github/workflows/ci.yaml
ericpromislow: Good catch

.github/workflows/ci.yaml
ericpromislow: Also I think we've discussed this in slack, but the `ubuntu:version` images are too bare-bones. I'd rather use an image that comes with the utilities we need (too bad about mockgen)

tomleb (approved):

nflynt (approved):

PR #427, opened 2024-05-01
Move to wrangler v3

tomleb:
# Issue 

With the [move to k8s 1.30](https://github.com/rancher/wrangler/pull/425) we had to do some breaking changes. This requires a new major version.


MbolotSuse (approved):
LGTM.

pkg/codegen/main.go
```diff
@@ -19,7 +19,7 @@ import (
 
 func main() {
 	controllergen.Run(args.Options{
-		ImportPackage: "github.com/rancher/wrangler/v2/pkg/generated",
+		ImportPackage: "github.com/rancher/wrangler/v3/pkg/generated",
 		OutputPackage: "github.com/rancher/wrangler/pkg/generated",
```
maxsokolovsky, line 23: Does this need to be updated?

tomleb: No it's not needed. Output package ends up being the path where the generated files are written to. For example: `/tmp/gopath/src/github.com/wrangler/pkg/generated`. Then whatever is under `/tmp/gopath/src/github.com/rancher/wrangler` is copied back to the current working directory. If that path had `v2/` in it, it would mess up the destination of the generated files.

That's why we both have ImportPackage, which does contain the v2 suffix, and OutputPackage, which doesn't.

ericpromislow (approved):
Verified that all the added `v3`s are due to deleted `v2`s, and that there are no remaining `v2`s.

PR #453, opened 2024-06-17
Migrate release-2.0-untagged to gha

ericpromislow:
Related to: [#45226](https://github.com/rancher/rancher/issues/45226) - for branch `release-2.0-untagged`


tomleb (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  ci:
+    strategy:
+      matrix:
+        arch:
+        - amd64
+        - arm64
+    runs-on: org-${{ github.repository_owner_id }}-${{ matrix.arch }}-k8s
+    container: registry.suse.com/bci/golang:1.22
```
nflynt, line 20: Shouldn't the go version in CI match the go version used by this branch? It seems like we have GHA using 1.22 while the base version should be 1.20 according to `go.mod` in _this_ branch: https://github.com/rancher/wrangler/blob/release-2.0-untagged/go.mod

nflynt (changes requested):

.github/workflows/ci.yaml
ericpromislow: Bumped the go version to 1.22 instead -- just because a branch is based on an older version doesn't mean it has to use an older version of go.

tomleb: For wrangler  & lasso we try to keep the same Go version that's being used by the k8s library to not add more constraints for users of those libraries. [k8s 1.27](https://github.com/kubernetes/kubernetes/blob/release-1.27/go.mod#L9) requires a minimum version of Go 1.20 so we would keep it as is and instead change the container to use golang 1.20.

ericpromislow: OK, I pulled this branch back to golang 1.20

nflynt (approved):

PR #468, opened 2024-09-05
Switch from github.com/golang/mock to go.uber.org/mock

aruiz14:
## Issue: rancher/rancher#46972

- Adds `go:generate` commands for `pkg/generic/fake`, so there is no need to generate it manually anymore.
- Refresh generated files
- Update imports to the new dependency.
- `go mod tidy`


pkg/generic/fake/README.md
```diff
@@ -65,16 +65,9 @@ mock.EXPECT().List(gomock.Any()).Return(nil, nil)
 ```
 
 ## Fake Generation
-This package was generated with `mockgen -package fake -destination ./controller.go -source ../controller.go` and `mockgen -package fake -destination ./cache.go -source ../cache.go`
+This package was generated with `mockgen`, see [`generate.go`](./generate.go).
```
ericpromislow, line 68: Is it possible to include the command that was run to generate this file?

I believe this option is in the uber fork of this library

ericpromislow (changes requested):
Just a quick comment. The rest is fine.

pkg/generic/fake/README.md
aruiz14: The fork will automatically add a hint of the `mockgen` command used to generate it, but it's not really customizable.
Nonetheless, the only command needed must be `go generate` (just added it to the README). There is no need for users/developers to run `mockgen` command directly.

ericpromislow (approved):

moio (approved):
Switching to a maintained version is a good thing (tm)

approved

PR #472, opened 2024-09-11
 Extend OnRemove to accept WithCondition option

aruiz14:
### Needs
- #468 

Please use the [commits view](https://github.com/rancher/wrangler/pull/472/commits) while reviewing this.

### Description
This changes the signature for `OnRemove` to add an optional `OnRemoveOption`s list, which keeps backwards compatibility with existing code (with the exception of possible mocks, which will need to be regenerated).

The new `WithCondition` option accepts a `func(runtime.Object) bool` (defaults to `includeAll`) so that finalizers will only be applied to resources matching the condition.

### DIFF
--- a/pkg/generic/remove_options.go
+++ b/pkg/generic/remove_options.go
@@ -5,7 +5,8 @@ import "k8s.io/apimachinery/pkg/runtime"
 // OnRemoveOption allows configuring OnRemove handlers
 type OnRemoveOption func(*objectLifecycleAdapter)
 
-// WithCondition will make OnRemove handlers to ignore resources not matching the provided condition
+// WithCondition will make OnRemove handlers to ignore resources not matching the provided condition.
+// Successive WithCondition options will replace previous ones; combining conditions needs to be done in the function being passed.
 func WithCondition(f func(runtime.Object) bool) OnRemoveOption {
 	return func(o *objectLifecycleAdapter) {
 		o.condition = f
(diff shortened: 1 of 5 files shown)


pkg/generic/remove_options.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "k8s.io/apimachinery/pkg/runtime"
+
+// OnRemoveOption allows configuring OnRemove handlers
+type OnRemoveOption func(*objectLifecycleAdapter)
+
+// WithCondition will make OnRemove handlers to ignore resources not matching the provided condition
+func WithCondition(f func(runtime.Object) bool) OnRemoveOption {
+	return func(o *objectLifecycleAdapter) {
+		o.condition = f
```
moio, line 11: You might want to clarify that the latest `WithCondition` wins in case multiple are specified (one could think they are combined with AND)

moio (approved):
No big objections from my part. Good job

pkg/generic/remove_options.go
aruiz14: Thanks, amended the comment at 86560d9

MbolotSuse (changes requested):
Now that #468 has merged, would you mind rebasing on `master` to clean up the diff a bit?

Outside of that, have two comments on the implementation.

aruiz14:
Used alternative approach directly using `AddGenericHandler` https://github.com/rancher/rancher/pull/47078

PR #473, opened 2024-09-11
Add unit tests for OnRemove

aruiz14:
Extracted from: https://github.com/rancher/wrangler/pull/472


ericpromislow (changes requested):
Pretty sure this should be a PR against `main` now.

pkg/generic/remove_test.go
```diff
@@ -0,0 +1,119 @@
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("sync() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if tt.wantFunc != nil {
+				if err := tt.wantFunc(got); err != nil {
+					t.Errorf("sync() unexpected value = %v, err: %v", got, err)
+				}
+			}
+			if got, want := handlerCount, tt.wantHandlerCount; got != want {
```
ericpromislow, line 111: Why the need for assigning to `got` and `want`? Why not this simpler code:
```
			if handlerCount != tt.wantHandlerCount {
				t.Errorf("handlerCount = %v, want %v", handlerCount, tt.wantHandlerCount)
			}
			if updaterCount != tt.wantUpdaterCount {
				t.Errorf("updaterCount = %v, want %v", updaterCount, tt.wantUpdaterCount)
			}
```

ericpromislow (changes requested):
It looks like the code I mentioned was more complex at some point and no longer needs the temporary variables.

pkg/generic/remove_test.go
aruiz14: It's just convention, see https://github.com/golang/go/issues/44179 (and https://go.dev/wiki/TestComments).
Defining `got, want` allows using those scoped variables in the error message.
Nonetheless, it's true that this specific case is simple enough (and the expected value is already named `tt.want...`) to just use your suggestion.

ericpromislow (approved):
After the explanation I'm fine either way

moio (approved):

PR #474, opened 2024-09-24
Added OpenAPI Codegen

gehrkefc:
This is related to issue: [47007](https://github.com/rancher/rancher/issues/47007)


pkg/controller-gen/main.go
```diff
@@ -265,6 +276,56 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
...
 
+func generateOpenAPI(groups map[string]bool, customArgs *cgargs.CustomArgs) error {
+	if len(groups) == 0 {
+		return nil
+	}
+
+	openAPIArgs := oaargs.New()
+	openAPIArgs.OutputDir = filepath.Join(customArgs.OutputBase, customArgs.Options.OutputPackage, "openapi")
+	openAPIArgs.OutputFile = "zz_generated_openapi.go"
+	openAPIArgs.OutputPkg = customArgs.Options.OutputPackage
```
tomleb, line 287:
```suggestion
	openAPIArgs.OutputPkg = customArgs.Options.OutputPackage+"/openapi"
```

Otherwise the generated package name `package <name>` will be different than `openapi`.

pkg/controller-gen/main.go
```diff
@@ -265,6 +276,56 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
 	)
 }
 
+func generateOpenAPI(groups map[string]bool, customArgs *cgargs.CustomArgs) error {
```
tomleb, line 279: While trying out the changes here for types in rancher/rancher, I couldn't get it to generate anything.

The generator works only when the directive `// +k8s:openapi-gen=true` is in the `doc.go` file. Adding it manually won't work because another code generator overwrites this doc.go file. This can be added here: https://github.com/rancher/wrangler/blob/241bfe6cb00df18d82e08966e8eb0217461cec39/pkg/controller-gen/generators/client_generator.go#L96-L100.

Note: This ends up in many files which we don't need (doc.go, zz_generated_register.go, zz_generated_list_types.go). Not ideal, but it's already like this and we want to get rid of gengo anyway.

pkg/controller-gen/main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
+	"golang.org/x/exp/maps"
```
tomleb, line 11: The `"maps"` package already exists iirc, can we use that?

go.mod
```diff
@@ -22,8 +24,9 @@ require (
 	k8s.io/client-go v0.31.1
 	k8s.io/code-generator v0.31.1
 	k8s.io/gengo v0.0.0-20240826214909-a7b603a56eb7
-	k8s.io/gengo/v2 v2.0.0-20240228010128-51d4e06bde70
+	k8s.io/gengo/v2 v2.0.0-20240826214909-a7b603a56eb7
 	k8s.io/kube-aggregator v0.31.1
+	k8s.io/kube-openapi v0.0.0-20240903163716-9e1beecbcb38
```
tomleb, line 29: I'd suggest keeping the version closer to our targeted k8s.io dependency. For example for v0.31.1, they're using [k8s.io/kube-openapi v0.0.0-20240228011516-70dd3763d340](https://github.com/kubernetes/kubernetes/blob/v1.31.1/go.mod#L109). Unless this prevent the PR from working.

pkg/controller-gen/main.go
```diff
@@ -265,6 +276,56 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
...
+		if len(group.OpenAPIDependencies) > 0 {
+			for _, dep := range group.OpenAPIDependencies {
+				if _, found := inputDirs[dep]; !found {
+					inputDirs[dep] = true
+				}
+			}
+		}
+	}
+
+	logrus.Info(inputDirs)
```
tomleb, line 318: Leftover from debugging ?

pkg/controller-gen/main.go
gehrkefc: Ok, thank you!

pkg/controller-gen/main.go
gehrkefc: Just added! Thank you!

go.mod
gehrkefc: Yeah, this makes the PR to not to work, looks like it's something to gengo version.

pkg/controller-gen/main.go
gehrkefc: I'm sorry, just removed. Thank you!

pkg/controller-gen/main.go
gehrkefc: We've discussed that on slack. I use the method Keys from maps package. And at time of this go version it wasn't available at the standard "maps" package.

go.mod
```diff
@@ -7,12 +7,14 @@ toolchain go1.22.6
 require (
 	github.com/evanphx/json-patch v5.9.0+incompatible
 	github.com/ghodss/yaml v1.0.0
+	github.com/labstack/echo/v4 v4.12.0
```
tomleb, line 10: This dependency doesn't seem used, is it?

go.mod
tomleb: Can you elaborate? Downgrading the kube-openapi version to that one seems to work fine for me (while keeping gengo with the version you have)

tomleb (commented):
2 more comments but otherwise looks good and worked fine in rancher/rancher to generate OpenAPI documents of types.

go.mod
gehrkefc: Yeah, you're right, was just testing some stuff with it.

gehrkefc:
@tomleb You were right, I worked to adjust both gengo and kube-openapi to compatible version and it worked! Thank you!

tomleb (dismissed):

pkg/controller-gen/main.go
```diff
@@ -265,6 +276,54 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
...
+		if !groups[gv.Group] {
+			continue
+		}
+
+		if _, found := inputDirs[names[0].Package]; !found {
+			inputDirs[names[0].Package] = true
+		}
+
+		group := customArgs.Options.Groups[gv.Group]
+		if len(group.OpenAPIDependencies) > 0 {
```
maxsokolovsky, line 309: Nit: you don't need this check, you could just run the loop anyways.

pkg/controller-gen/main.go
```diff
@@ -265,6 +276,54 @@ func generateClientset(groups map[string]bool, customArgs *cgargs.CustomArgs) er
...
+	openAPIArgs.OutputPkg = customArgs.Options.OutputPackage + "/openapi"
+	openAPIArgs.GoHeaderFile = customArgs.Options.Boilerplate
+
+	if err := openAPIArgs.Validate(); err != nil {
+		return err
+	}
+
+	getTargets := func(context *generator.Context) []generator.Target {
+		return oa.GetTargets(context, openAPIArgs)
+	}
```
maxsokolovsky, line 296: Nit: I'd move this variable closer to where it's used below.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/stretchr/testify v1.9.0
 	go.uber.org/mock v0.4.0
+	golang.org/x/exp v0.0.0-20230515195305-f3d0a9c9a5cc
```
maxsokolovsky, line 16: Not sure if we want the extra dependency for a simple one-off function of getting keys from a map into a slice. If we had more use of the module, then I wouldn't think about it.

maxsokolovsky (commented):
Looks good! Had just a couple of very small questions...

pkg/controller-gen/main.go
gehrkefc: Thanks!

maxsokolovsky (approved):

pkg/controller-gen/generators/client_generator.go
```diff
@@ -95,6 +95,7 @@ func (cg *ClientGenerator) typesGroupVersionDocPackage(name *types.Name, gv sche
 	p.HeaderComment = []byte(fmt.Sprintf(`
 %s
 
+// +k8s:openapi-gen=true
```
tomleb, line 98: We should make this dependent on the GenerateOpenAPI flag.

 I thought we were okay with just doing the same as deepcopy but I misunderstood that deepcopy is generated when this generator runs..

It will also prevent having a huge amount of file changes in rancher/rancher when we update wrangler. (And also will prevent any change in behavior due to that new directive)

Could you make that change? Sorry about this.

pkg/controller-gen/main.go
```diff
@@ -117,6 +123,10 @@ func Run(opts cgargs.Options) {
 		logrus.Fatalf("informers failed: %v", err)
 	}
 
+	if err := generateOpenAPI(openAPIGroups, customArgs); err != nil {
+		logrus.Fatalf("openapi docs: %v", err)
```
ericpromislow, line 127: Shouldn't the error message be `"openapi failed: %v", err` ?

Also, unless I'm missing something, all the `"...%v", err)` instances in the `Run` function here should be using `%w` instead. This should probably be a separate TechDebt issue, but you can start at this line.

gehrkefc: Ok, agreed! Do you want me to  change others as well ?

gehrkefc: Ok, agreed!

ericpromislow: Sure. I don't know if there are tests hitting that code in `rancher/rancher`, but it's standard to use `%w` on errors, so why not.

ericpromislow (changes requested):
Just a couple of cosmetic changes related to errors

pkg/controller-gen/main.go
gehrkefc: logrus.Fatalf doesn't support %w, so I had to revert it to %v again.

ericpromislow (approved):
LGTM

tomleb (approved):
Tested it in r/r and code with the flag off don't modify current generated code and when the flag is on it does generate properly.

PR #487, opened 2024-11-22
Set namespace field when creating relationships for helm releases

### LINKED ISSUE #486: Resources listing for Helm charts breaks if the chart refers to another namespace
I was originally going to open this issue at rancher/dashboard, but I started going through the code and realized the problem was actually here.

The summarizer for apps.catalog.cattle.io is not properly setting the namespace of deployed Helm resources. When the relationships are processed by steve, it fallbacks to the using namespace of the release, since there's nothing set on the summary.

Below are extra info that I was originally gonna report to rancher/dashboard:

**Describe the bug**
On the Resources tab of Installed App, the UI is ignoring the namespace of the resources deployed with Helm and instead lists the release namespace for all of them. This not only results in incorrect display, but also when clicking in a given resource it uses the wrong namespace as well, resulting in an error page.

**Expected Result**
The interface should list the correct namespaces for each deployed resource and the links should also point to the right namespace.

**Screenshots**
In this screenshot, both ecr-credentials-updater-aws-registry-kube-system and ecr-credentials-updater-aws-registry-test-namespace RoleBindings belong to kube-system and test-namespace, respectively. But the UI wrongly points to av-infra (the namespace that the chart is installed to).

image

When clicking on one of them, it results in an error, since its trying to get the resource from the wrong namespace:
image

Here's a few screenshots from the API viewer:

At spec.resources, all entries have their namespace properly set:
image

But the metadata.relationships used by the UI is wrong - all resources are prefixed with av-infra, instead of the correct namespaces:
image

cienijr:
Fixes #486

### DIFF
--- a/pkg/summary/cattletypes.go
+++ b/pkg/summary/cattletypes.go
@@ -34,6 +34,7 @@ func checkRelease(obj data.Object, _ []Condition, summary Summary) Summary {
 	for _, resources := range obj.Slice("spec", "resources") {
 		summary.Relationships = append(summary.Relationships, Relationship{
 			Name:       resources.String("name"),
+			Namespace:  resources.String("namespace"),
 			Kind:       resources.String("kind"),
 			APIVersion: resources.String("apiVersion"),
 			Type:       "helmresource",


crobby (dismissed):
@cienijr I believe this should be against the main branch rather than master.  That is a change we made fairly recently.  Thanks

cienijr:
@crobby thanks for the help. I initially opened it against the main branch, but changed it after seeing a few other PRs in this repo. I reverted it back.

gehrkefc (approved):
LGTM

crobby (approved):
seems reasonable to me

cienijr:
Hello @crobby @gehrkefc 😄

Thank you both for reviewing and approving my PR. I'm not at all familiar with the Rancher development process and just wanted to check if there's anything else I should be doing to help land these changes. Please let me know if I can assist further 🚀

cienijr:
Hey there,

I see that no actions were ever run for my PR. Is this the blocker for getting this change upstream?
If so, what extra steps are required to run these checks?

image

PR #491, opened 2024-12-02
#47207 - Updated to go v1.23 and Changed Deepcopy to use gengo/v2

gehrkefc:


brandond:
1. We need this merged for k3s and rke2 1.32.
2. Upstream made a mess of gengo; as of https://github.com/kubernetes/gengo/pull/277 it no longer prunes imports but will instead helpfully `gofmt` your whole project when you run it. Discussion at: https://github.com/kubernetes/gengo/pull/277#issuecomment-2556653931 and I have a fork at https://github.com/brandond/gengo with the affected code reverted. We may need to cary this fork in Rancher if upstream does not revert that change or at least make the behavior configurable.

gehrkefc:
This work was moved to https://github.com/rancher/wrangler/pull/492

PR #492, opened 2024-12-20
Update to golang 1.23.3 / k8s 1.32

brandond:
Builds on changes from
* https://github.com/rancher/wrangler/pull/491
* https://github.com/kubernetes/gengo/pull/277#issuecomment-2557462569


pkg/controller-gen/main.go
```diff
@@ -61,6 +63,15 @@ func Run(opts cgargs.Options) {
...
 	getTargets := func(context *generator.Context) []generator.Target {
+		// replace the default formatter options to ensure unused imports are pruned.
+		// ref: https://github.com/kubernetes/gengo/pull/277#issuecomment-2557462569
+		if fileType, ok := context.FileTypes["go"]; ok {
+			if defaultFileType, ok := fileType.(*generator.DefaultFileType); ok {
+				defaultFileType.Format = func(src []byte) ([]byte, error) {
+					return imports.Process("", src, nil)
+				}
+			}
+		}
```
tomleb, line 74: I think that'd be safer. He's also mentioned using a `sync.Once` which I think we should, though I haven't digged into the gengo code to see why that would require it.

```suggestion
               goGenerator := generators.NewGoFile()
               goGenerator.Format = func(src []byte) ([]byte, error) {
					return imports.Process("", src, nil)
				}
				context.FileType[generators.GoFileType] = goGenerator
```

brandond: We could wrap it in a sync.once but I added debug printfs to both this getTargets func and the custom Format function and it appears the context is reused through all the rest of the Execute calls so it's not necessary to replace it more than once. It also seems harmless TO replace it more than once though.

brandond: I've confirmed that this also works for k3s

pkg/controller-gen/main.go
```diff
@@ -200,25 +211,44 @@ func generateDeepcopy(groups map[string]bool, customArgs *cgargs.CustomArgs) err
...
 		if !groups[gv.Group] {
 			continue
 		}
-		args.InputDirs = append(args.InputDirs, names[0].Package)
-		deepCopyCustomArgs.BoundingDirs = append(deepCopyCustomArgs.BoundingDirs, names[0].Package)
+		inputDirs = append(inputDirs, names[0].Package)
+		deepCopyArgs.BoundingDirs = append(deepCopyArgs.BoundingDirs, names[0].Package)
+	}
+
+	getTargets := func(context *generator.Context) []generator.Target {
```
brandond, line 227: I don't remember changing this bit

pkg/controller-gen/main.go
tomleb: Great. What do you think of using the suggested approach instead of doing a type check of an interface? That way this won't silently break when we update gengo.

Basically, instead of getting the default file generator for "go" filetype, let's just create it ourselves and modify its `Format` field.

Also, `"go"` -> generators.GoFileType.

tomleb (approved):

pkg/controller-gen/main.go
```diff
@@ -200,25 +209,44 @@ func generateDeepcopy(groups map[string]bool, customArgs *cgargs.CustomArgs) err
...
+				PkgName:        t.Name(),
+				PkgPath:        t.Path(),
+				PkgDir:         t.Dir(),
+				HeaderComment:  t.Header(deepCopyArgs.OutputFile),
+				FilterFunc:     t.Filter,
+				GeneratorsFunc: t.Generators,
+			})
+		}
+
+		return newTargets
```
gehrkefc, line 240: You can just return the dp.GetTargets(...) from here if you want. I just did it because I had the need of change the dir at that point. But I'll approve it anyway.

gehrkefc (approved):

pkg/controller-gen/main.go
brandond: You're suggesting dropping the target modifications into newTargets and just returning dp.GetTargets directly?

pkg/controller-gen/main.go
brandond: changed

tomleb (approved):

PR #506, opened 2025-01-13
Adjust go.mod values for go/toolchain

mallardduck:
When updating BRO for 1.32, I found that the current config caused the CI to be brittle as it required a specific GO 1.23 release. Specifically the error I was encountering is:

>  go.mod requires go >= 1.23.3 (running go 1.23.2; GOTOOLCHAIN=local)

This change should allow for `1.23.2` (and older 1.23) to work without errors.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/rancher/wrangler/v3
 
-go 1.23.3
+go 1.23.0
+
+toolchain go1.23.4
 
 require (
 	github.com/evanphx/json-patch v5.9.0+incompatible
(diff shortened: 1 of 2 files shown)


tomleb (approved):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/wrangler/v3
 
-go 1.23.3
+go 1.23.0
+
+toolchain go1.23.4
```
brandond, line 5: This should be fine... we're on toolchain 1.23.3 for 1.32 at the moment but I suspect the next upstream release will bump that.

On the distros side we always use the same toolchain version as upstream, for any given patch release of Kubernetes:
https://github.com/kubernetes/kubernetes/blob/v1.32.0/build/dependencies.yaml#L118-L119

brandond (approved):

PR #514, opened 2025-01-29
Add override for MicroTime to "date" 

### LINKED ISSUE #513: Wrangler CRD schema generation incorrectly identifies MicroTime as object in resulting openapi spec
Scenario:

You have a new type, say `Foo`. Defined as such:

```go
type Foo struct {
... // objectMeta, typeMeta
EventTime *metav1.MicroTime `json:"eventTime"`
}
```

Now, you run this type through Wrangler's provided `ToOpenAPIFromStruct` as seen in https://github.com/rancher/wrangler/blob/0bfb013edb606f656cb209a75396defec9fa4038/pkg/schemas/openapi/generate.go#L25

Wrangler _incorrectly_ spits out an OpenAPI schema in which the MicroTime is stored as an object. It *should* be stored as a string, similar to how metav1.Time is stored. 

The problem is in reflection.go, https://github.com/rancher/wrangler/blob/0bfb013edb606f656cb209a75396defec9fa4038/pkg/schemas/reflection.go#L495-L498

Time is correctly compensated for, but not MicroTime.

ebauman:
This PR adds a type override for "MicroTime" fields, correctly identifying them as "date" for later OpenAPI spec gen.

This fixes #513

### DIFF
--- a/pkg/schemas/reflection.go
+++ b/pkg/schemas/reflection.go
@@ -496,6 +496,9 @@ func (s *Schemas) determineSchemaType(t reflect.Type) (string, error) {
 		if t.Name() == "Time" {
 			return "date", nil
 		}
+		if t.Name() == "MicroTime" {
+			return "date", nil
+		}
 		if t.Name() == "IntOrString" {
 			return "intOrString", nil
 		}


ericpromislow:
We don't have a current place to unit-test this, do we?  It looks ok otherwise.

PR #563, opened 2025-04-04
#48559 - Added GVK check errors

gehrkefc:
# Issue: https://github.com/rancher/rancher/issues/48559

### DIFF
--- a/pkg/summary/gvk.go
+++ b/pkg/summary/gvk.go
@@ -2,7 +2,8 @@ package summary
 
 import (
 	"encoding/json"
-	"errors"
+	"fmt"
+	"regexp"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -10,22 +11,26 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, 
-// Versions and Kinds to Conditions, Types, and Status, indicating with a flag if a certain condition with specific 
+var (
+	gvkRegExp = regexp.MustCompile(`^(.*?)/(.*),\s*Kind=(.+)$`)
+)
+
+// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups,
+// Versions and Kinds to Conditions, Types, and Status, indicating with a flag if a certain condition with specific
 // status represents an error or not. It is expected to be something like:
 //
-//{
-//	"gvk": 			"helm.cattle.io/v1, Kind=HelmChart",
-//	"conditionMappings": [
-//		{
-//			"type": "JobCreated"	// This means JobCreated is mostly informational and True or False 
-//		},				// doesn't mean error
-//		{
-//			"type": "Failed",	// This means Failed is considered error if it's status is True
-//			"status": ["True"]
-//		},
+//	{
+//		"gvk": 			"helm.cattle.io/v1, Kind=HelmChart",
+//		"conditionMappings": [
+//			{
+//				"type": "JobCreated"	// This means JobCreated is mostly informational and True or False
+//			},				// doesn't mean error
+//			{
+//				"type": "Failed",	// This means Failed is considered error if it's status is True
+//				"status": ["True"]
+//			},
+//		}
 //	}
-//}
 type conditionTypeStatusJSON struct {
 	GVK               string                     `json:"gvk"`
 	ConditionMappings []conditionStatusErrorJSON `json:"conditionMappings"`
@@ -61,23 +66,23 @@ func (m ConditionTypeStatusErrorMapping) UnmarshalJSON(data []byte) error {
 	}
 
 	for _, mapping := range conditionMappingsJSON {
+		// checking if mapping.GVK is in the right format: <group>/[version], Kind=[kind]
+		mx := gvkRegExp.FindStringSubmatch(mapping.GVK)
+		if len(mx) == 0 {
+			return fmt.Errorf("gvk parsing failed: wrong GVK format: <%s>\n", mapping.GVK)
+		}
+
 		// parsing Group, Version and Kind in a format of group/version, Kind=kind
 		// eg: helm.cattle.io/v1, Kind=HelmChart, becomes:
 		// parts[0] = helm.cattle.io/v1
 		// parts[1] = HelmChart
 		parts := strings.Split(mapping.GVK, ", Kind=")
-		if len(parts) != 2 {
-			return errors.New("gvk parsing failed: wrong GVK format")
-		}
 
 		// parsing group/version
 		// eg: helm.cattle.io/v1, becomes:
 		// gv[0] = helm.cattle.io
 		// gv[1] = v1
 		gv := strings.Split(parts[0], "/")
-		if len(gv) != 2 {
-			return errors.New("gvk parsing failed: wrong group/version format")
-		}
 
 		conditionMappings := map[string]sets.Set[metav1.ConditionStatus]{}
 		for _, condition := range mapping.ConditionMappings {
--- a/pkg/summary/gvk.go
+++ b/pkg/summary/gvk.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
-	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -72,27 +71,15 @@ func (m ConditionTypeStatusErrorMapping) UnmarshalJSON(data []byte) error {
 			return fmt.Errorf("gvk parsing failed: wrong GVK format: <%s>\n", mapping.GVK)
 		}
 
-		// parsing Group, Version and Kind in a format of group/version, Kind=kind
-		// eg: helm.cattle.io/v1, Kind=HelmChart, becomes:
-		// parts[0] = helm.cattle.io/v1
-		// parts[1] = HelmChart
-		parts := strings.Split(mapping.GVK, ", Kind=")
-
-		// parsing group/version
-		// eg: helm.cattle.io/v1, becomes:
-		// gv[0] = helm.cattle.io
-		// gv[1] = v1
-		gv := strings.Split(parts[0], "/")
-
 		conditionMappings := map[string]sets.Set[metav1.ConditionStatus]{}
 		for _, condition := range mapping.ConditionMappings {
 			conditionMappings[condition.Type] = sets.New[metav1.ConditionStatus](condition.Status...)
 		}
 
 		m[schema.GroupVersionKind{
-			Group:   gv[0],
-			Version: gv[1],
-			Kind:    parts[1],
+			Group:   mx[1],
+			Version: mx[2],
+			Kind:    mx[3],
 		}] = conditionMappings
 	}
 	return nil
--- a/pkg/summary/gvk.go
+++ b/pkg/summary/gvk.go
@@ -10,22 +10,22 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, Versions and Kinds to
-// Conditions, Types, and Status, indicating with a flag if a certain condition with specific status represents an error or not.
-// It is expected to be something like:
+// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, 
+// Versions and Kinds to Conditions, Types, and Status, indicating with a flag if a certain condition with specific 
+// status represents an error or not. It is expected to be something like:
 //
-//	{
-//			"gvk": 			"helm.cattle.io/v1, Kind=HelmChart",
-//			"conditionMappings": [
-//				{
-//					"type": "JobCreated"	// This means JobCreated is mostly informational and True or False doesn't mean error
-//				},
-//				{
-//					"type": "Failed",	// This means Failed is considered error if it's status is True
-//					"status": ["True"]
-//				},
-//			}
+//{
+//	"gvk": 			"helm.cattle.io/v1, Kind=HelmChart",
+//	"conditionMappings": [
+//		{
+//			"type": "JobCreated"	// This means JobCreated is mostly informational and True or False 
+//		},				// doesn't mean error
+//		{
+//			"type": "Failed",	// This means Failed is considered error if it's status is True
+//			"status": ["True"]
+//		},
 //	}
+//}
 type conditionTypeStatusJSON struct {
 	GVK               string                     `json:"gvk"`
 	ConditionMappings []conditionStatusErrorJSON `json:"conditionMappings"`
--- a/pkg/summary/gvk.go
+++ b/pkg/summary/gvk.go
@@ -16,8 +16,8 @@ const (
 )
 
 // conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, Versions and Kinds to
-// Conditions Types and Status, indicating with a flag if a certain condition with specific status represents an error or not.
-// Is is expected to be something like:
+// Conditions, Types, and Status, indicating with a flag if a certain condition with specific status represents an error or not.
+// It is expected to be something like:
 //
 //	{
 //			"gvk": 			"helm.cattle.io/v1/HelmChart",
@@ -32,7 +32,7 @@ const (
 //			}
 //	}
 //
-// IMPORTANT: Please pay attention to the the conditionStatusSep char, in this case it is a '/'. It separates Groups, Versions and Kinds.
+// IMPORTANT: Please pay attention to the conditionStatusSep char, in this case it is a '/'. It separates Groups, Versions and Kinds.
 type conditionTypeStatusJSON struct {
 	GVK              string                     `json:"gvk"`
 	ConditionMapping []conditionStatusErrorJSON `json:"conditionMapping"`
--- a/pkg/summary/gvk.go
+++ b/pkg/summary/gvk.go
@@ -1,0 +1,115 @@
+package summary
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	conditionStatusSep = "/"
+	gvkFormat          = "%s" + conditionStatusSep + "%s" + conditionStatusSep + "%s"
+)
+
+// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, Versions and Kinds to
+// Conditions Types and Status, indicating with a flag if a certain condition with specific status represents an error or not.
+// Is is expected to be something like:
+//
+//	{
+//			"gvk": 				"helm.cattle.io/v1/HelmChart",
+//			"conditionMapping": [
+//				{
+//					"type": "JobCreated"
+//					"status": "True",
+//					"error": false,
+//				},
+//				{
+//					"type": "Failed"
+//					"status": "True",
+//					"error": true,
+//				},
+//				{
+//					"type": "Failed"
+//					"status": "False",
+//					"error": false,
+//				},
+//			}
+//	}
+//
+// IMPORTANT: Please pay attention to the the conditionStatusSep char, in this case it is a '/'. It separates Groups, Versions and Kinds.
+type conditionTypeStatusJSON struct {
+	GVK              string                     `json:"gvk"`
+	ConditionMapping []conditionStatusErrorJSON `json:"conditionMapping"`
+}
+
+type conditionStatusErrorJSON struct {
+	Type    string `json:"type"`
+	Status  string `json:"status"`
+	IsError bool   `json:"error"`
+}
+
+type ConditionTypeStatus struct {
+	Type   string
+	Status metav1.ConditionStatus
+}
+
+type ConditionTypeStatusErrorMapping map[schema.GroupVersionKind]map[ConditionTypeStatus]bool
+
+func (m ConditionTypeStatusErrorMapping) MarshalJSON() ([]byte, error) {
+	output := []conditionTypeStatusJSON{}
+	for gvk, mapping := range m {
+		output = append(output, conditionTypeStatusJSON{
+			GVK: fmt.Sprintf(gvkFormat, gvk.Group, gvk.Version, gvk.Kind),
+			ConditionMapping: func() []conditionStatusErrorJSON {
+				conditions := make([]conditionStatusErrorJSON, 0)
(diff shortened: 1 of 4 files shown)


pkg/summary/gvk.go
```diff
@@ -0,0 +1,115 @@
...
+type conditionTypeStatusJSON struct {
+	GVK              string                     `json:"gvk"`
+	ConditionMapping []conditionStatusErrorJSON `json:"conditionMapping"`
+}
+
+type conditionStatusErrorJSON struct {
+	Type    string `json:"type"`
+	Status  string `json:"status"`
+	IsError bool   `json:"error"`
+}
```
brandond, line 53:
```suggestion
type conditionStatusErrorJSON struct {
	Type    string `json:"type"`
	Conditions  []string `json:"conditions,omitempty"`
}
```

pkg/summary/gvk.go
```diff
@@ -0,0 +1,115 @@
...
+//					"type": "Failed"
+//					"status": "True",
+//					"error": true,
+//				},
+//				{
+//					"type": "Failed"
+//					"status": "False",
+//					"error": false,
+//				},
+//			}
```
brandond, line 40: I think that you could be more concise with this if you used only a single item for each type, and simply stored a list of condition status strings (if any) that are considered errors. An empty list means the type is purely informational and there are no errors.
```suggestion
//				{
//					"type": "JobCreated",
//				},
//				{
//					"type": "Failed"
//					"conditions": ["True"],
//				},
//			}
```

pkg/summary/summarizers.go
```diff
@@ -82,6 +88,18 @@ var (
...
 
+	// For given GVK, This condition Type and this Status, indicates an error or not
+	// e.g.: GVK: helm.cattle.io/v1, HelmChart
+	//            --> Type: JobCreated, Status: True, IsError: false
+	//            --> Type: Failed, Status: True, IsError: false
+	GVKConditionErrorMapping = ConditionTypeStatusErrorMapping{
+		{Group: "helm.cattle.io", Version: "v1", Kind: "HelmChart"}: {
+			{"JobCreated", metav1.ConditionTrue}: false,
+			{"Failed", metav1.ConditionFalse}:    false,
+			{"Failed", metav1.ConditionTrue}:     true,
```
brandond, line 99: It shouldn't be necessary to list the "Failed" type twice here, with "False" => false and "True" => true as the error states. Map each type once, with a list of states that are errors.
```suggestion
			{"JobCreated", nil},
			{"Failed", []string{metav1.ConditionTrue}},
```

brandond (changes requested):
I would rework this to track a list of error conditions for each type, instead of having to repeat each type+condition pair.

Use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets
```golang
sets.New[string](mapping.Conditions).Has(condition)
```

crobby (commented):
Is there an issue to map this PR to?

brandond (approved):
one question on the input validation, lgtm otherwise!

gehrkefc:
> Is there an issue to map this PR to?

Yeah, I'm sorry. Just updated it.

pkg/summary/gvk.go
```diff
@@ -0,0 +1,107 @@
...
+		gvk := strings.Split(mapping.GVK, conditionStatusSep)
+		if len(gvk) != 3 {
+			return errors.New("gvk parsing failed: wrong GVK format")
+		}
+		conditionMapping := map[string]sets.Set[string]{}
+		for _, condition := range mapping.ConditionMapping {
+			// checking if statuses are "True" or "False"
+			for _, status := range condition.Status {
+				if status != string(metav1.ConditionFalse) && status != string(metav1.ConditionTrue) {
+					return errors.New("gvk parsing failed: conditions status should be only True or False")
```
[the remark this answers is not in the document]
gehrkefc: @brandond Ok, I'll remove that.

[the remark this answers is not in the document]
gehrkefc: @brandond Thank you!

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+						Type: condition,
+						Status: func() []string {
+							output := []string{}
+							for status := range statuses {
+								output = append(output, status)
+							}
+							return output
+						}(),
+					})
+				}
```
brandond, line 66: I'm curious what is going on here, why is all this necessary? This is just a struct with one `string` and one `[]string` field. It should marshal properly without all this extra logic.

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	conditionStatusSep = "/"
+	gvkFormat          = "%s" + conditionStatusSep + "%s" + conditionStatusSep + "%s"
```
brandond, line 15: This isn't separating the status, it's separating the GVK parts.

Also, we shouldn't invent a new format for stringifying GVKs. GVKs already have a `String()` method that formats them in a preferred format that we should reuse: `helm.cattle.io/v1, Kind=HelmChart`

See the helpers at https://github.com/k3s-io/k3s/blob/c2efae3e1cf4a3e761ec297edd542504d7166b46/pkg/deploy/controller.go#L507-L524

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, Versions and Kinds to
+// Conditions Types and Status, indicating with a flag if a certain condition with specific status represents an error or not.
+// Is is expected to be something like:
```
ericpromislow, line 20: Typo: should be `It is`

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	conditionStatusSep = "/"
+	gvkFormat          = "%s" + conditionStatusSep + "%s" + conditionStatusSep + "%s"
+)
+
+// conditionTypeStatusJSON is a custom JSON to map a complex object into a standard JSON object. It maps Groups, Versions and Kinds to
+// Conditions Types and Status, indicating with a flag if a certain condition with specific status represents an error or not.
```
ericpromislow, line 19: style nit: `Conditions, Types, and Status`

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+//					"type": "JobCreated"	// This means JobCreated is mostly informational and True or False doesn't mean error
+//				},
+//				{
+//					"type": "Failed",	// This means Failed is considered error if it's status is True
+//					"status": ["True"]
+//				},
+//			}
+//	}
+//
+// IMPORTANT: Please pay attention to the the conditionStatusSep char, in this case it is a '/'. It separates Groups, Versions and Kinds.
```
ericpromislow, line 35: typo: `Please pay attention to the conditionStatusSep`

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+type conditionTypeStatusJSON struct {
+	GVK              string                     `json:"gvk"`
+	ConditionMapping []conditionStatusErrorJSON `json:"conditionMapping"`
+}
+
+type conditionStatusErrorJSON struct {
+	Type   string   `json:"type"`
+	Status []string `json:"status"`
```
brandond, line 43:
```suggestion
	Status []metav1.ConditionStatus `json:"status,omitempty"`
```

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+							for status := range statuses {
+								output = append(output, status)
+							}
+							return output
+						}(),
+					})
+				}
+				return conditions
+			}(),
+		})
```
brandond, line 69: You can do this in much fewer lines without the nested funcs: 

```suggestion
    typeStatus := conditionTypeStatusJSON{GVK: gvk.String()}
    for condition, statuses := range mapping {
      typeStatus.ConditionMapping = append(typeStatus.ConditionMapping, conditionStatusErrorJSON{Type: condition, Status: statuses.UnsortedList()})
    }
    output = append(output, typeStatus)
```

pkg/summary/gvk.go
```diff
@@ -0,0 +1,100 @@
...
+	GVK              string                     `json:"gvk"`
+	ConditionMapping []conditionStatusErrorJSON `json:"conditionMapping"`
+}
+
+type conditionStatusErrorJSON struct {
+	Type   string   `json:"type"`
+	Status []string `json:"status"`
+}
+
+type ConditionTypeStatusErrorMapping map[schema.GroupVersionKind]map[string]sets.Set[string]
```
brandond, line 46:
```suggestion
type ConditionTypeStatusErrorMapping map[schema.GroupVersionKind]map[string]sets.Set[metav1.ConditionStatus]
```

pkg/summary/summarizers.go
```diff
@@ -82,6 +89,19 @@ var (
...
+	// For given GVK, This condition Type and this Status, indicates an error or not
+	// e.g.: GVK: helm.cattle.io/v1, HelmChart
+	//            --> Type: JobCreated: [], indicates True or False are not errors
+	//            --> Type: Failed: ["True"], indicates "True" status is considered error
+	// 	      --> Type: Worked: ["False"], indicates "False" status is considered error
+	// 	      --> Type: Unknown: ["True", "False"] indicated "True" or "False" are considered errors
+	GVKConditionErrorMapping = ConditionTypeStatusErrorMapping{
+		{Group: "helm.cattle.io", Version: "v1", Kind: "HelmChart"}: {
+			"JobCreated": sets.New[string](),
+			"Failed":     sets.New[string](string(metav1.ConditionTrue)),
```
brandond, line 101:
```suggestion
			"Failed":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```

brandond (changes requested):
I think we could still reduce the complexity here by a lot, to support something like this:

```json
[
  {
    "gvk": "helm.cattle.io/v1, Kind=HelmChart",
    "conditionMapping": [
      {
        "type": "JobCreated"
      },
      {
        "type": "Failed",
        "status": [
          "True"
        ]
      }
    ]
  }
]
```

pkg/summary/gvk.go
gehrkefc: Fixed it. Thanks!

pkg/summary/gvk.go
gehrkefc: Agreed on all. Thanks!

pkg/summary/summarizers.go
```diff
@@ -74,14 +81,26 @@ var (
 		"ReplicaFailure":      true,
 		"Stalled":             true,
 	}
```
brandond, line 83: Should this list here be reworked to use the same logic? Doesn't feel right to have a separate legacy list of node status error mappings, now that we have a proper way to map resource status conditions to errors.

pkg/summary/gvk.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (m ConditionTypeStatusErrorMapping) UnmarshalJSON(data []byte) error {
+	var conditionMappingsJSON []conditionTypeStatusJSON
+	err := json.Unmarshal(data, &conditionMappingsJSON)
+	if err != nil {
+		return err
+	}
+
+	for _, mapping := range conditionMappingsJSON {
+		// parsing Group, Version and Kind in a format of group/version, Kind=kind
```
ericpromislow, line 64: I think lines 64 through 80 can be replaced with a regex, set before the loop 
```
rx := regexp.MustCompile(`^(.*?)/(.*),\s*Kind=(.+)$`)
```

and 
```
m := rx.FindStringSubmatch(mapping.GVK)
if len(m) == 0 {
  return fmt.Errorf("not in GVK format: <%s>\n", mapping.GVK)
}
	// fmt.Printf("G: %s, V: %s, K: %s\n", m[1], m[2], m[3])

File: pkg/summary/gvk.go [hunk 16]
```diff
@@ -0,0 +1,94 @@
...
+
+type conditionStatusErrorJSON struct {
+	Type   string                   `json:"type"`
+	Status []metav1.ConditionStatus `json:"status,omitempty"`
+}
+
+type ConditionTypeStatusErrorMapping map[schema.GroupVersionKind]map[string]sets.Set[metav1.ConditionStatus]
+
+func (m ConditionTypeStatusErrorMapping) MarshalJSON() ([]byte, error) {
+	output := []conditionTypeStatusJSON{}
```
ericpromislow, line 42: There must be a better name for the variable you're using to gather the json than `output`. Maybe `statuses`?  `output` implies something is getting emitted somewhere.

pkg/summary/summarizers.go
```diff
@@ -218,7 +260,60 @@ func checkStandard(obj data.Object, _ []Condition, summary Summary) Summary {
...
+	}
+
+	obj, err := json.Marshal(data)
+	if err != nil {
+		return summary, false
+	}
+
+	gvk, detected, err := gvk.Detect(obj)
+	if err != nil {
+		return summary, false
```
ericpromislow, line 275: Same here about discarding an error situation

ericpromislow (changes requested):
Getting closer, more understandable too people new to this code (like me). Just some style tweak requests

pkg/summary/gvk.go
brandond: > output implies something is getting emitted somewhere.

That is exactly what's happening. `conditionTypeStatusJSON` is the intermediate type used for translation between the internal maps and the JSON structure. That structure is what is actually marshalled and output.

pkg/summary/gvk.go
```diff
@@ -0,0 +1,99 @@
...
+		// eg: helm.cattle.io/v1, Kind=HelmChart, becomes:
+		// parts[0] = helm.cattle.io/v1
+		// parts[1] = HelmChart
+		parts := strings.Split(mapping.GVK, ", Kind=")
+
+		// parsing group/version
+		// eg: helm.cattle.io/v1, becomes:
+		// gv[0] = helm.cattle.io
+		// gv[1] = v1
+		gv := strings.Split(parts[0], "/")
```
brandond, line 85: are you still going to rework this to use the regex submatches, instead of splitting? I think that was intended to be the whole point of the regex.

pkg/summary/gvk.go
```diff
@@ -0,0 +1,99 @@
...
+
+		conditionMappings := map[string]sets.Set[metav1.ConditionStatus]{}
+		for _, condition := range mapping.ConditionMappings {
+			conditionMappings[condition.Type] = sets.New[metav1.ConditionStatus](condition.Status...)
+		}
+
+		m[schema.GroupVersionKind{
+			Group:   gv[0],
+			Version: gv[1],
+			Kind:    parts[1],
```
brandond, line 95: just use the captured portions from the regex match

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,40 @@ var (
...
+	//		--> Unknown: ["True", "False"] indicated "True" or "False" are considered errors
+	GVKConditionErrorMapping = ConditionTypeStatusErrorMapping{
+		{Group: "helm.cattle.io", Version: "v1", Kind: "HelmChart"}: {
+			"JobCreated": sets.New[metav1.ConditionStatus](),
+			"Failed":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		},
+		{Group: "", Version: "v1", Kind: "Node"}: {
+			"MemoryPressure":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"DiskPressure":       sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"NetworkUnavailable": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```
brandond, line 86: This should have all of
```
		"OutOfDisk":          sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
		"MemoryPressure":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
		"DiskPressure":       sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
		"NetworkUnavailable": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
		"Unschedulable":      sets.New[metav1.ConditionStatus](metav1.ConditionTrue),

```

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,40 @@ var (
...
+		},
+		{Group: "apps", Version: "v1", Kind: "Deployment"}: {
+			"ReplicaFailure": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"Stalled":        sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```
brandond, line 90: Remove Stalled.

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,40 @@ var (
...
+			"JobCreated": sets.New[metav1.ConditionStatus](),
+			"Failed":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		},
+		{Group: "", Version: "v1", Kind: "Node"}: {
+			"MemoryPressure":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"DiskPressure":       sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"NetworkUnavailable": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		},
+		{Group: "apps", Version: "v1", Kind: "Deployment"}: {
+			"ReplicaFailure": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```
brandond, line 89: ReplicaFailure also needs to be an error for `apps/v1 Kind=ReplicaSet`

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,40 @@ var (
...
+	// mapping.
+	FallbackConditionErrorMapping = map[string]sets.Set[metav1.ConditionStatus]{
+		"OutOfDisk":          sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"MemoryPressure":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"DiskPressure":       sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"NetworkUnavailable": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"Unschedulable":      sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"ReplicaFailure":     sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"Stalled":            sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		"Failed":             sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```
brandond, line 105: This should only contain Stalled, as all the other condition types are expected to be found on specific Kinds. Stalled can be found on any CR Kind managed via a Rancher controller that uses the rancher kstatus helpers:
https://github.com/rancher/wrangler/blob/master/pkg/kstatus/kstatus.go

pkg/summary/gvk.go
gehrkefc: yeah, will do! Thanks!

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,38 @@ var (
...
+		},
+		{Group: "apps", Version: "v1", Kind: "ReplicaSet"}: {
+			"ReplicaFailure": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		},
+	}
+
+	// In case we cannot match any Groups, Versions and Kinds in GVKConditionErrorMapping then we fallback to this
+	// mapping.
+	FallbackConditionErrorMapping = map[string]sets.Set[metav1.ConditionStatus]{
+		"Unschedulable":      sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```
brandond, line 101: I think we can drop this one as `Unschedulable` is a not a condition Type, it is a Reason on the `PodScheduled` condition on Pods. It would not be matching anything at the moment.

gehrkefc: Ok. Will remove that. Just noticed a flaky test. Will work to fix it also.

ericpromislow (changes requested):
Changes look good, but CI is failing.

When I run `scripts/ci` locally,I get this error:
```
Validating
Running: golangci-lint
pkg/summary/summarizers.go:101: File is not `goimports`-ed (goimports)
                "Stalled":            sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
                "Failed":             sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
```

pkg/summary/summarizers.go
```diff
@@ -61,25 +69,37 @@ var (
...
+			"ReplicaFailure": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+			"Progressing":    sets.New[metav1.ConditionStatus](metav1.ConditionFalse),
+		},
+		{Group: "apps", Version: "v1", Kind: "ReplicaSet"}: {
+			"ReplicaFailure": sets.New[metav1.ConditionStatus](metav1.ConditionTrue),
+		},
+	}
+
+	// In case we cannot match any Groups, Versions and Kinds in GVKConditionErrorMapping then we fallback to this
+	// mapping.
```
brandond, line 99: Would it be simpler to store this in GVKConditionErrorMapping as an empty GVK? This would also allow for updating it via the env var by specifying an empty string for the GVK. The json unmarshal code would need to allow for that.

pkg/summary/summarizers.go
```diff
@@ -218,28 +289,47 @@ func checkStandard(obj data.Object, _ []Condition, summary Summary) Summary {
...
-		}
+func checkErrors(data data.Object, conditions []Condition, summary Summary) Summary {
+	conditionMapping := FallbackConditionErrorMapping
+
+	if len(conditions) == 0 {
+		return summary
 	}
 
-	if summary.Error {
+	obj, err := json.Marshal(data)
```
brandond, line 299: We shouldn't be marshalling this to get the GVK, that's expensive. Do as checkOwner does up above, and wrap this in `unstructured.Unstructured` and use the helpers on that type to get the GVK.

pkg/summary/summarizers.go
```diff
@@ -218,28 +289,47 @@ func checkStandard(obj data.Object, _ []Condition, summary Summary) Summary {
...
+	if err != nil {
+		logrus.Debugln("checkErrors: failed to detect GVK: ", err.Error())
+	}
+
+	if detected {
+		gvkConditionMapping, found := GVKConditionErrorMapping[gvk]
+		if found {
+			conditionMapping = gvkConditionMapping
+		}
+	}
```
brandond, line 315: You could simplify this using `GVKConditionErrorMapping[gvk]` if found, and GVKConditionErrorMapping[schema.GroupVersionKind{}] for the fallback if not found.

pkg/summary/summarizers.go
gehrkefc: I was looking for an alternative. Will use that one then Thanks!

brandond (approved):

ericpromislow (approved):
Looks good

PR #566, opened 2025-05-01
Fix and tighten up 'GetValueFromAny'

ericpromislow:
I discovered this while working on issue [#42767](https://github.com/rancher/rancher/issues/42767) -- steve failed to sort on a value in a string array in non-sql mode, and then I saw that `GetValueFromAny` here in wrangler wasn't handling string arrays, and needed some improvement (like missing `else-blocks` to handle unexpected cases, duplicated handling of the end of a list.

There will be a steve PR once this is in.

### DIFF
--- a/pkg/data/values.go
+++ b/pkg/data/values.go
@@ -41,41 +41,46 @@ func GetValue(data map[string]interface{}, keys ...string) (interface{}, bool) {
 	return nil, false
 }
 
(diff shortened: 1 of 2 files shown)


crobby (approved):
Sorry, I mean to get to this yesterday.
I think this looks good.  Existing tests all work and you've added new tests.  lgtm

gehrkefc (approved):

PR #570, opened 2025-05-09
#48559 - Added Reason Check

gehrkefc:
This is adding back a Reason check removed at: https://github.com/rancher/wrangler/pull/563

Issue: https://github.com/rancher/rancher/issues/48559

### DIFF
--- a/pkg/summary/summarizers.go
+++ b/pkg/summary/summarizers.go
@@ -304,11 +304,13 @@ func checkErrors(data data.Object, conditions []Condition, summary Summary) Summ
 
 	for _, c := range conditions {
 		status, found := conditionMapping[c.Type()]
-		if !found {
+		reasonIsError := c.Reason() == "Error"
+
+		if !found && !reasonIsError {
 			continue
 		}
 
-		if status.Has(metav1.ConditionStatus(c.Status())) || c.Reason() == "Error" {
+		if status.Has(metav1.ConditionStatus(c.Status())) || reasonIsError {
 			summary.Error = true
 			summary.Message = append(summary.Message, c.Message())
 			if summary.State == "active" || summary.State == "" {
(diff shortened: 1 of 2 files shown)


pkg/summary/summarizers.go
```diff
@@ -308,7 +308,7 @@ func checkErrors(data data.Object, conditions []Condition, summary Summary) Summ
 			continue
 		}
 
-		if status.Has(metav1.ConditionStatus(c.Status())) {
+		if status.Has(metav1.ConditionStatus(c.Status())) || c.Reason() == "Error" {
```
brandond, line 311: This needs to be above the `status, found := conditionMapping[c.Type()]` check on line 306. `c.Reason() == "Error"` indicates an error regardless of whether or not the condition type is present in the map.

brandond (changes requested):

pkg/summary/summarizers.go
gehrkefc: Sure, just addressed it. Also fixed the test. Thank you!

pkg/summary/summarizers.go
```diff
@@ -304,11 +304,13 @@ func checkErrors(data data.Object, conditions []Condition, summary Summary) Summ
...
 		status, found := conditionMapping[c.Type()]
-		if !found {
+		reasonIsError := c.Reason() == "Error"
+
+		if !found && !reasonIsError {
 			continue
 		}
 
-		if status.Has(metav1.ConditionStatus(c.Status())) {
+		if status.Has(metav1.ConditionStatus(c.Status())) || reasonIsError {
```
brandond, line 313: nit: I'd probably put the `reasonIsError` test first to short-circuit checking `status.Has` if we already know it's an error.
```suggestion
		if reasonIsError || status.Has(metav1.ConditionStatus(c.Status())) {
```

brandond (changes requested):
one nit on the test order

brandond (approved):

ericpromislow (approved):
LGTM

PR #573, opened 2025-06-10
[v1.33] Bump dependencies for v1.33

krunalhinguu:
### Issues

- https://github.com/rancher/rancher/issues/48796


rohitsakala (approved):

crobby:
@krunalhinguu 
I'm late to the party here, but we should also update the VERSIONS file in this repo to update the supported k8s version.

PR #579, opened 2025-07-10
Add checkGeneration method for checking CRDS with controllers

rohitsakala:
Issue : https://github.com/rancher/rancher/issues/49671

- More details in this PR : https://github.com/rancher/steve/pull/714
- Adding a `checkGeneration` function which compares `ObservedGeneration` and `metadata.Generation` to determine the status of the resource. It uses the `hasObservedGeneration` value passed on by `steve`.

Steve PR https://github.com/rancher/steve/pull/714

Looking for feedback on the solution

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/moby/locker v1.0.1
 	github.com/pkg/errors v0.9.1
-	github.com/rancher/lasso v0.2.3-rc3
+	github.com/rancher/lasso v0.2.3
 	github.com/sirupsen/logrus v1.9.3
 	github.com/stretchr/testify v1.10.0
 	go.uber.org/mock v0.5.2
--- a/go.sum
+++ b/go.sum
@@ -114,8 +114,8 @@ github.com/prometheus/common v0.62.0 h1:xasJaQlnWAeyHdUBeGjXmutelfJHWMRr+Fg4QszZ
 github.com/prometheus/common v0.62.0/go.mod h1:vyBcEuLSvWos9B1+CyL7JZ2up+uFzXhkqml0W5zIY1I=
 github.com/prometheus/procfs v0.15.1 h1:YagwOFzUgYfKKHX6Dr+sHT7km/hxC76UB0learggepc=
 github.com/prometheus/procfs v0.15.1/go.mod h1:fB45yRUv8NstnjriLhBQLuOUt+WW4BsoGhij/e3PBqk=
-github.com/rancher/lasso v0.2.3-rc3 h1:kkYnARdFeY6A9E2XnjfQbG8CssHQwobPMIFqPRGpVxc=
-github.com/rancher/lasso v0.2.3-rc3/go.mod h1:G+KeeOaKRjp+qGp0bV6VbLhYrq1vHbJPbDh40ejg5yE=
+github.com/rancher/lasso v0.2.3 h1:74/z/C/O3ykhyMrRuEgc9kVyYiSoS7kp5BAijlcyXDg=
+github.com/rancher/lasso v0.2.3/go.mod h1:G+KeeOaKRjp+qGp0bV6VbLhYrq1vHbJPbDh40ejg5yE=
 github.com/rogpeppe/go-internal v1.13.1 h1:KvO1DLK/DRN07sQ1LQKScxyZJuNnedQ5/wKSR38lUII=
 github.com/rogpeppe/go-internal v1.13.1/go.mod h1:uMEvuHeurkdAXX61udpOXGD/AzZDWNMNyH2VO9fmH0o=
 github.com/russross/blackfriday/v2 v2.1.0/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=


pkg/summary/summarizers.go
```diff
@@ -204,6 +206,22 @@ func initializeCheckErrors() {
 	logrus.Debugf("GVK Error Mapping not provided, using predefined values")
 }
 
+func checkGeneration(obj data.Object, _ []Condition, summary Summary) Summary {
+	if summary.Attributes["hasObservedGeneration"] == true {
```
tomleb: Do we need this attribute? Can't we look at .status.observedGeneration from the obj directly to know if it has one? eg: if it's set, then it has one, if not, it doesn't?

pkg/summary/summarizers.go
```diff
@@ -135,6 +136,7 @@ func init() {
 	}
 
 	Summarizers = []Summarizer{
+		checkGeneration,
```
tomleb: I thought the [issue pointed at in this comment](https://github.com/rancher/rancher/issues/49671#issuecomment-2864284813) was that there was no `.status` since the object was created. How does looking at observedGeneration and metadata generation solves this?

I understand that observedGeneration can be used to know if the status is stale by comparing metadata generation.

pkg/summary/summarizers.go
rohitsuse: No, when the object is just created, it won't have any status even though the CRD has status and ObservedGeneration field defined. When the controller has processed it, the controller will then set the `status` subresource.

tomleb: >No, when the object is just created, it won't have any status even though the CRD has status and ObservedGeneration field defined. 

Right, so can't we detect that case from the object directly, set status to in-progress (or unknown, I'm not sure what the states we have are).

Then, when the status is added, our controllers should receive that information and we'll just recalculate the summary ?

pkg/summary/summarizers.go
rohitsuse: The controller adds the status https://github.com/rancher/rancher/blob/main/pkg/controllers/dashboard/helm/repo.go#L386 and returns the object to `controller framework` and it updates the obj as whatever is returned from the controller (the controller adds/updates the status). 

So until the controller acts on it, the resource is in `Inactive` state and should show as `Inactive` in the UI but the kstatus reports as `Active` which is incorrect https://github.com/kubernetes-sigs/cli-utils/issues/632

pkg/summary/summarizers.go
rohitsuse: Sorry can you elaborate on this ? 

How can I detect that case from the object directly ? the obj is something like an instance of the CRD. It has no informtion about the definition. 

I think I might have missed something here.

tomleb: Something like:

```go
	_, hasObservedGeneration := data.GetValue(obj, "status", "observedGeneration")

```

`data` module is from wrangler (pkg/data). (obj is a `map[string]any`)

rohitsuse: It is still returning `false` 

Here is what I tested. 

Screenshot 2025-07-21 at 1 45 30 PM


Screenshot 2025-07-21 at 1 45 49 PM

In fact, ClusterRepo CRD has `ObservedGeneration`. 

Note: Once the controller has processed this, the above code will return true becuase, the controller will put the status after it processes it.

tomleb: Discussed offline, we need this because we'll get misleading data if we check the generation but the object doesn't have an observedgeneration field.

pkg/summary/informer/informer.go
```diff
@@ -116,6 +117,39 @@ func (f *summarySharedInformerFactory) WaitForCacheSync(stopCh <-chan struct{})
...
 
+func NewFilteredSummaryInformerWithSchema(
+	client client.ExtendedInterface,
+	gvr schema.GroupVersionResource,
+	schema *schemas.Schema,
+	namespace string,
+	resyncPeriod time.Duration,
+	indexers cache.Indexers,
+	tweakListOptions TweakListOptionsFunc,
+) informers.GenericInformer {
```
ericpromislow, line 127: From what I can tell currently there are no callers of this function, so it's hard to see what kinds of `tweakListOptions` funcs are being called.  But the problem I have with it is that it invokes `tweakListOptions(&options)` twice, once in the ListFunc and once in the WatchFunc.  Couldn't you do something like:
```
func NewFil...(...) {
  if tweakListOptions != nil {
    tweakListOptions(&options)
  }
  return 
  ...
  &cache.ListWatch{
				ListFunc: func(…) (…) {
					return client.ResourceWithSchema(gvr, schema).Namespace(namespace).List(context.TODO(), options)
				},
   ```
   
   No need to tweak the options structure each time?

Comment (rohitsuse) [reply pkg/summary/informer/informer.go:127]: This function will be called in steve and rancher in my PRs.

Comment (rohitsuse) [reply pkg/summary/informer/informer.go:127]: We need to call tweatOptoons(&options) two times ince `&options` is an argument sent by the List and watch separately.

File: pkg/summary/summarizers.go [hunk 4]
```diff
@@ -204,6 +206,22 @@ func initializeCheckErrors() {
 	logrus.Debugf("GVK Error Mapping not provided, using predefined values")
 }
 
+func checkGeneration(obj data.Object, _ []Condition, summary Summary) Summary {
+	if summary.Attributes != nil && summary.Attributes["hasObservedGeneration"] == true {
+		metadataGenerationStr := obj.String("metadata", "generation")
+		observedGenerationStr := obj.String("status", "observedGeneration")
```
tomleb: We could add `obj.ToNumber()` perhaps?

pkg/summary/summarizers.go
```diff
@@ -204,6 +206,22 @@ func initializeCheckErrors() {
...
+		generation, errMetadata := strconv.ParseInt(metadataGenerationStr, 10, 64)
+		observed, _ := strconv.ParseInt(observedGenerationStr, 10, 64)
```
tomleb: We want a unparsed observedGeneration to mean in progress, correct?

pkg/summary/summarizers.go
```diff
@@ -204,6 +206,22 @@ func initializeCheckErrors() {
...
+
+		if errMetadata == nil && observed < generation {
```
tomleb: Should we treat observed > generation to also mean in progress? Or active? (this would happen only if a controller has a bug though)

pkg/summary/summary.go
```diff
@@ -109,6 +117,12 @@ func Summarize(runtimeObj runtime.Object) Summary {
 	}
 
 	conditions := getConditions(obj)
+	if opts != nil {
+		if summary.Attributes == nil {
+			summary.Attributes = make(map[string]interface{})
```
tomleb: nit: Not a fan of using summary.Attributes for this instead of a proper typed field.. Have you considered adding a new field (unexported perhaps) to summary?

tomleb (commented):
Some questions but overall seems to do the job.

pkg/summary/summarizers.go
rohitsuse: This case will be handled by the kstatus library. since kstatus library also checks observedgeneration and metadata.generation.

rohitsuse: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/status/generic.go#L73

pkg/summary/summarizers.go
rohitsuse: Right. Let me fix this.

pkg/summary/summarizers.go
rohitsuse: Kstatus uses `!=` . So I would the same here.

pkg/summary/summary.go
rohitsuse: No I didn't. Summary.Attributes is not anywhere. 

I could use typed field but though does Summary struct really need to have direct infromation about a schema field ?

But thinking now, it is might not be a bad idea.

pkg/summary/summarizers.go
rohitsuse: I am using the `apimachinery` library which give feels like more appropriate like in here https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/status/generic.go#L73.

that package is better than ours.

pkg/summary/summarizers.go
rohitsuse: Fixed. Similar to https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/status/generic.go#L73

pkg/summary/summary.go
rohitsuse: Fixed it.

pkg/summary/summary.go
```diff
@@ -11,12 +11,17 @@ import (
...
-	Message       []string               `json:"message,omitempty"`
-	Attributes    map[string]interface{} `json:"-"`
-	Relationships []Relationship         `json:"-"`
+	State                 string                 `json:"state,omitempty"`
+	Error                 bool                   `json:"error,omitempty"`
+	Transitioning         bool                   `json:"transitioning,omitempty"`
+	Message               []string               `json:"message,omitempty"`
+	Attributes            map[string]interface{} `json:"-"`
+	Relationships         []Relationship         `json:"-"`
+	HasObservedGeneration bool                   `json:"hasObservedGeneration,omitempty"`
```
tomleb: We should avoid making this available to UI for now unless they ask for it. Let's keep it internal only (with `json:"-"`.

pkg/summary/summary.go
tomleb: Resolving this thread but will open a new one.

pkg/summary/client/simple.go
```diff
@@ -113,3 +130,19 @@ type watcher struct {
 func (w watcher) ResultChan() <-chan watch.Event {
 	return w.eventChan
 }
+
+func generateSummarizeOpts(schema *schemas.Schema) *summary.SummarizeOptions {
+	hasObservedGeneration := false
+	if schema != nil && schema.Attributes != nil {
+		if v, ok := schema.Attributes["hasObservedGeneration"]; ok {
```
tomleb, line 137: Where is this set? I couldn't find it in steve and couldn't find it in this PR :thinking: 

I'm also wondering if we should use the schema at all. I kinda hate the idea of having arbitrary attributes as `map[string]any`. Thought?

pkg/summary/summarizers_test.go
```diff
@@ -364,3 +364,74 @@ func TestCheckErrors(t *testing.T) {
...
+				"status": map[string]interface{}{
+					"observedGeneration": int64(2),
+				},
+			},
+			summary:   Summary{HasObservedGeneration: true, State: "foo"},
+			wantState: "foo",
+			wantTrans: false,
+		},
+		{
+			name: "observedGeneration does not equal generation, sets in-progress and transitioning",
```
tomleb, line 444: We probably want some cases for non-int64 values here just for completeness. I'm also wondering, will this fail if the generation is an int32 or just int instead of an int64? Should it?

pkg/summary/summary.go
rohitsuse: Done.

pkg/summary/client/simple.go
rohitsuse: - It is set here https://github.com/rancher/steve/pull/714/files#diff-dba2dcfa1b1ec304ae9cbd72a09b386ec299d157964c3aa1d96d7be5d633b76eR82.
- I think we should have schema since logically schema has a one to one mapping to a CRDefinitiion.
- I can change HasObservedGeneration from attributes to a new field. Let me know.

pkg/summary/summarizers_test.go
rohitsuse: I would like to follow the upstream. It only works for [int64](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/status/generic.go#L73). So this should as well. 

Added additional testcases.

pkg/summary/summarizers.go
```diff
@@ -204,6 +205,36 @@ func initializeCheckErrors() {
...
+func checkGeneration(obj data.Object, _ []Condition, summary Summary) Summary {
+	if summary.State != "" {
+		return summary
+	}
+
+	if summary.HasObservedGeneration {
+		metadataGeneration, metadataFound, errMetadata := unstructured.NestedInt64(obj, "metadata", "generation")
+		if !metadataFound {
+			return summary
+		}
```
tomleb: Error should be checked before using any other values. Right now this check is eating errors.

pkg/summary/client/simple.go
tomleb: Ok thanks for showing me that this attributes is in schema which is embedded in types.APISchema.

This Attributes is only used in steve so far (afaik) and it has a bunch of methods to interact with it.. I'm wondering if we should make this a new field or have the functions to access / set those fields be in wrangler instead.. Becaues otherwise we have `hasObservedGeneration` magic var here but it's only set in another codebase..

pkg/summary/summarizers.go
rohitsuse: Yes switched places. Fixed it.

pkg/summary/client/simple.go
rohitsuse: Moved the set and access methods to wrangler.

pkg/summary/summarizers_test.go
```diff
@@ -364,3 +364,103 @@ func TestCheckErrors(t *testing.T) {
...
+
+func TestCheckGeneration(t *testing.T) {
+	tests := []struct {
+		name      string
+		obj       data.Object
+		summary   Summary
+		wantState string
+		wantTrans bool
+	}{
+		{
```
tomleb, line 376: Can we add a test for when .status doesn't yet exist? Let's make sure it return in-progress, right?

rohitsuse: Yes.

rohitsuse: Added.Done.

ericpromislow (approved):
Looks fine.

PR #582, opened 2025-07-30
47898 OnLeader should return any error returned by the called functions.

marytlf:
### **Problem**
The Rancher system experienced panics during startup due to a non-idempotent OnLeader handler. The existing code included a retry mechanism (a for loop) that repeatedly executed this handler. Since the handler was not designed to be idempotent (meaning it produces the same result regardless of how many times it's run), retrying it after an error exacerbated the issue, leading to crashes rather than graceful error handling. The core problem was the mismatch between the handler's non-idempotent nature and the code's assumption of idempotency through its retry logic.

### **Solution**
The proposed solution addresses the problem by removing the retry mechanism. Instead of the for loop that re-attempts the f(m.leaderCTX) call, the fix executes the leader function only once. If an error occurs during this single execution, it's logged using logrus.Errorf. If the function executes successfully, a success message is logged using logrus.Infof. This change ensures that non-idempotent operations are not retried, allowing errors to be immediately visible in the logs without causing further panics due to repeated execution of the faulty handler.

###**Tests**

- Leader Re-election (if applicable to Rancher's leader election):

**Test:** If Rancher's leader election process can involve a leader stepping down and a new one being elected, test this scenario to ensure the OnLeader handler is correctly invoked for the new leader and the fix still holds.

**Expected Outcome:** The handler should execute once for the new leader, logging success or failure appropriately without panics or retries.

- Concurrency/Stress Test:

**Test:** Simulate high load during Rancher startup to stress the leader election and handler execution.

**Expected Outcome:** The system should remain stable, and any errors should be logged gracefully without leading to crashes, confirming the removal of the retry mechanism prevents cascading failures.

### DIFF
--- a/pkg/leader/manager.go
+++ b/pkg/leader/manager.go
@@ -3,7 +3,6 @@ package leader
 import (
 	"context"
 	"sync"
-	"time"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
@@ -47,13 +46,10 @@ func (m *Manager) Start(ctx context.Context) {
 func (m *Manager) OnLeader(f func(ctx context.Context) error) {
 	go func() {
 		<-m.leaderChan
-		for {
-			if err := f(m.leaderCTX); err != nil {
-				logrus.Errorf("failed to call leader func: %v", err)
-				time.Sleep(5 * time.Second)
-				continue
-			}
-			break
+		if err := f(m.leaderCTX); err != nil {
+			logrus.Errorf("failed to call leader func: %v", err)
+		} else {
+			logrus.Infof("leader func executed successfully")
 		}
 	}()
 }


pkg/leader/manager.go
```diff
@@ -47,13 +46,10 @@ func (m *Manager) Start(ctx context.Context) {
...
 		<-m.leaderChan
-		for {
-			if err := f(m.leaderCTX); err != nil {
-				logrus.Errorf("failed to call leader func: %v", err)
-				time.Sleep(5 * time.Second)
-				continue
-			}
-			break
+		if err := f(m.leaderCTX); err != nil {
+			logrus.Errorf("failed to call leader func: %v", err)
```
tomleb: From Rancher's perspective, we'll want to crash if there's an issue in most cases. So I think something like logrus.Fatal (if it exists) should be ran instead.

pkg/leader/manager.go
```diff
@@ -47,13 +46,10 @@ func (m *Manager) Start(ctx context.Context) {
...
+		} else {
+			logrus.Infof("leader func executed successfully")
```
tomleb: That sounds like a useful log message in general. However, it lacks context. Which leader func has executed successfully? And similarly to the failure case, which one has failed?

Maybe we want to introduce an identifier somewhere to distinguish between those.

tomleb (commented):
Wrangler is a bit different than other repositories maintained by frameworks: it's used by more than just Rancher, like k3s, and rke2. We gotta be careful how we make changes to it to try not to break those other projects.

So changing `OnLeader` which retries to the same version that doesn't retry could cause issues. It doesn't really, at least not for k3s/rke2 because I checked that they're not using it. In any case, it's good to try to keep old behavior working while still trying to make changes.

One way to achieve our goal while still making our changes would be to introduce a new function. For example,

```go
func (m *Manager) OnLeaderOrDie(f func(ctx context.Context) error {
  // This one doesn't retry and quits when it fails
})
```

This way, old code are still doing retry, and code in Rancher where we want to crash will need to be updated to use OnLeaderOrDie. (Note: The name I picked is because we already have something `RunOrDie` in that module, so seemed fitting).

Fwiw that's not the only way, we could also be more flexible by doing:

```go
type Options struct {
  Retry bool
  // We can add new things that changes the behavior
}

func (m *Manager) OnLeaderWithOptions(f func(ctx context.Context, opts Options) error {
  // This one doesn't retry and quits when it fails
})
```

It depends on how much we think this will require changes in the future. Kinda hard to predict. I think in this case, we're okay with just adding an OrDie variant, and if a third use case comes up in the future, we might introduce the more "flexible" option above.

pkg/leader/manager.go
```diff
@@ -43,6 +43,18 @@ func (m *Manager) Start(ctx context.Context) {
 	})
 }
 
+// OnLeaderOrDie this function will be called when leadership is acquired or die if failed
+func (m *Manager) OnLeaderOrDie(f func(ctx context.Context) error) {
+	go func() {
+		<-m.leaderChan
+		if err := f(m.leaderCTX); err != nil {
+			logrus.Fatalf("leader func: %s failed be executed :: %v", m.name, err)
```
tomleb: m.name is the name of the manager, right?

We'll have many OnLeaderOrDie() call calling many leader func. The manager name will be the same for each of them, so it doesn't give us enough information to distinguish between leader func.

Also nit (but more idiomatic error format):
```suggestion
			logrus.Fatalf("leader func: %s failed be executed: %v", m.name, err)
```

tomleb: So perhaps we want to have:

```go
func (m *Manager) OnLeaderOrDie(name string, f func(ctx context.Context) error) {
```
?

What do you think?

pkg/leader/manager.go
```diff
@@ -43,6 +43,18 @@ func (m *Manager) Start(ctx context.Context) {
 	})
 }
 
+// OnLeaderOrDie this function will be called when leadership is acquired or die if failed
+func (m *Manager) OnLeaderOrDie(name string, f func(ctx context.Context) error) {
+	go func() {
+		<-m.leaderChan
+		if err := f(m.leaderCTX); err != nil {
+			logrus.Fatalf("%s leader func failed be executed: %v", name, err)
+		} else {
```
bigkevmcd, line 52: You don't need the `else` here because `logrus.Fatalf` will have terminated the process so the other branch will never be executed?

bigkevmcd (approved):

PR #584, opened 2025-08-18
Fix potential race with cert generation and adding unit tests for needacert functionality

crobby:
Issue: https://github.com/rancher/rancher/issues/44584

### DIFF
--- a/pkg/needacert/needacert_test.go
+++ b/pkg/needacert/needacert_test.go
@@ -576,7 +576,7 @@ func TestHandler_GenerateSecret_Race(t *testing.T) {
 		secrets:      mockSecrets,
 	}
 
-	const concurrency = 50
+	const concurrency = 10
 	done := make(chan struct{})
 	for i := 0; i < concurrency; i++ {
 		go func() {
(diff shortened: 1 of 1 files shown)


pkg/needacert/needacert_test.go
```diff
@@ -0,0 +1,914 @@
...
+				},
+			},
+		},
+	}
+
+	updated, err := h.OnMutationWebhookChange("key", webhook)
+	assert.NoError(t, err)
+	assert.NotNil(t, updated)
+	assert.NotEmpty(t, updated.Webhooks[0].ClientConfig.CABundle)
+	assert.True(t, bytes.HasPrefix(updated.Webhooks[0].ClientConfig.CABundle, []byte("-----BEGIN CERTIFICATE-----")))
```
joshmeranda, line 174: Not a blocker, but would it be better to use `tls.X509KeyPair(updated.webhooks[0].ClientConfig.CABundle, keyPEM)` here to validate that it is a valid cert with the right values?

joshmeranda (approved):

pkg/needacert/needacert.go
```diff
@@ -273,19 +274,23 @@ func (h *handler) generateSecret(service *corev1.Service) (*corev1.Secret, error
 	dnsNames := dnsNameSet.List()
 	secret, err := h.secretsCache.Get(service.Namespace, secretName)
```
rohitsuse, line 275: We are already checking here for getting the secret ? Why are we again checking in line 283 ? 

Why can't we just directly lookup into the secrets client instead of the cache client always if we believe that there is a chance that cache is invalid.

Note: This is a non blocker comment.

rohitsakala (approved):

PR #594, opened 2025-10-02
Add core/v1 LimitRange and ResourceQuota generated controllers

aruiz14:
Seems to be needed for rancher/rancher#52103


andreas-kupries (approved):
LGTM, and thanks for the speed.

PR #595, opened 2025-10-06
Fix needacert overly updating webhook resources

anmazzotti:
This PR fixes needacert updating ValidatingWebhookConfigurations and MutatingWebhookConfiguration **per-webhook**.

In some extreme cases, like for [capz](https://github.com/kubernetes-sigs/cluster-api-provider-azure/releases/download/v1.21.0/infrastructure-components.yaml) `azureserviceoperator`, there can be as many as 350 webhooks.

With this change the resources are updated all at once, and only if they need updates. This also saves unnecessary writes when there are no changes to be made, assuming the Update operation is a PUT and not a PATCH one.

### DIFF
--- a/pkg/needacert/needacert.go
+++ b/pkg/needacert/needacert.go
@@ -6,7 +6,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


anmazzotti:
Sorry, I just validated this locally and it does still generate an increasingly amount of resource generations.
Put the PR in draft until I can find out.

anmazzotti:
Nevermind, the remaining issue seem to be with Turtles. This PR works as expected and can be reviewed. Thank you.

crobby (commented):
Looks pretty straightforward, but ideally we would add a unit test to go along with this.

rohitsakala (approved):

PR #600, opened 2025-10-11
Add secret handler to trigger needacert handlers when secret is rotated manually and when cert expires

rohitsakala:
Issue : https://github.com/rancher/rancher/issues/52097

### DIFF
--- a/pkg/needacert/needacert.go
+++ b/pkg/needacert/needacert.go
@@ -313,34 +313,42 @@ func (h *handler) generateSecret(service *corev1.Service) (*corev1.Secret, error
 		if err != nil {
 			return nil, err
 		}
-		created, err := h.secrets.Create(newSecret)
+		secret, err = h.secrets.Create(newSecret)
 		if apierror.IsAlreadyExists(err) {
-			return h.secrets.Get(service.Namespace, secretName, metav1.GetOptions{})
+			secret, err = h.secrets.Get(service.Namespace, secretName, metav1.GetOptions{})
+			if err != nil {
+				return nil, err
+			}
+		} else if err != nil {
+			return nil, err
 		}
-		return created, err
 	} else if err != nil {
 		return nil, err
 	}
 
-	if updated, err := h.updateSecret(service, secret, dnsNames); err != nil {
-		return nil, err
+	cert, parseErr := parseCert(secret)
+	if parseErr != nil {
+		return nil, parseErr
+	}
+
+	updated, updateErr := h.updateSecret(service, secret, dnsNames, cert)
+	if updateErr != nil {
+		return nil, updateErr
 	} else if updated != nil {
-		return h.secrets.Update(updated)
+		secret, err = h.secrets.Update(updated)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err := h.scheduleNextCertCheck(service, secret); err != nil {
+		return nil, fmt.Errorf("failed to schedule next cert check: %w", err)
 	}
 
 	return secret, nil
 }
 
-func (h *handler) updateSecret(owner runtime.Object, secret *corev1.Secret, dnsNames []string) (*corev1.Secret, error) {
-	tlsCert, err := tls.X509KeyPair(secret.Data[corev1.TLSCertKey], secret.Data[corev1.TLSPrivateKeyKey])
-	if err != nil || len(tlsCert.Certificate) == 0 {
-		return nil, err
-	}
-
-	cert, err := x509.ParseCertificate(tlsCert.Certificate[0])
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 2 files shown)


pkg/needacert/needacert.go
```diff
@@ -59,13 +60,35 @@ func Register(ctx context.Context,
...
+	secrets.OnChange(ctx, "need-a-cert", h.OnSecretChange)
+}
+
+// OnSecretChange handles Secret changes and enqueues the related Service.
+func (h *handler) OnSecretChange(key string, secret *corev1.Secret) (*corev1.Secret, error) {
+	if secret == nil {
+		return secret, nil
+	}
+
+	services, err := h.services.List(secret.Namespace, metav1.ListOptions{})
```
crobby, line 72: Should this use the serviceCache?

crobby (commented):
Looks pretty good.  Just one question about maybe using the serviceCache

pkg/needacert/needacert.go
gehrkefc: Only thing I could think of after reviewing it too.

rohitsuse: I used service since this trigger is called very rarely only when secret is rotated manually which not many do. So I used service. Let me know your thoughts.

crobby (approved):
lgtm

pkg/needacert/needacert.go
```diff
@@ -309,15 +345,20 @@ func (h *handler) generateSecret(service *corev1.Service) (*corev1.Secret, error
 }
 
 func (h *handler) updateSecret(owner runtime.Object, secret *corev1.Secret, dnsNames []string) (*corev1.Secret, error) {
-	tlsCert, err := tls.X509KeyPair(secret.Data[corev1.TLSCertKey], secret.Data[corev1.TLSPrivateKeyKey])
-	if err != nil || len(tlsCert.Certificate) == 0 {
+	cert, err := parseCert(secret)
```
moio: Optonal: is this call here necessary?

I see `parseCert` being called here and then again as the first thing inside of `scheduleNextCertCheck` few lines below.

I wonder if it would read better to only do the parsing inside of `scheduleNextCertCheck`.

_Note: this is a style remark and totally optional to address, you can feel free to ignore it_

pkg/needacert/needacert.go
```diff
@@ -292,9 +315,22 @@ func (h *handler) generateSecret(service *corev1.Service) (*corev1.Secret, error
 		}
 		created, err := h.secrets.Create(newSecret)
 		if apierror.IsAlreadyExists(err) {
-			return h.secrets.Get(service.Namespace, secretName, metav1.GetOptions{})
+			existing, getErr := h.secrets.Get(service.Namespace, secretName, metav1.GetOptions{})
+			if getErr != nil {
+				return nil, getErr
+			}
+			if err := h.scheduleNextCertCheck(service, existing, nil); err != nil {
```
moio: Optional: one thing I found a bit tricky to follow is that `scheduleNextChertCheck` is called in all cases: if the secret creation succeeds, if it fails because of a create/create conflict, _and_ if the secret already existed in the first place (because `updateSecret` calls `scheduleNextChertCheck`).

One possibility would be to keep  `updateSecret` and `scheduleNextChertCheck` separated, so in this method it's visible that `scheduleNextChertCheck` is called in all cases.

_Note: this is a style remark and totally optional to address, you can feel free to ignore it_

pkg/needacert/needacert.go
rohitsuse: We need it in the updateSecret function. Check the if condition in the updateSecret.

pkg/needacert/needacert.go
rohitsuse: I agree it is a bit confusing to read. Maybe I will try to add it in a defer function ?

gehrkefc (approved):
LGTM

pkg/needacert/needacert.go
```diff
@@ -59,13 +60,35 @@ func Register(ctx context.Context,
 	validatingController.OnChange(ctx, "need-a-cert", h.OnValidatingWebhookChange)
 	crdController.OnChange(ctx, "need-a-cert", h.OnCRDChange)
 	service.OnChange(ctx, "need-a-cert", h.OnService)
+	secrets.OnChange(ctx, "need-a-cert", h.OnSecretChange)
+}
+
+// OnSecretChange handles Secret changes and enqueues the related Service.
+func (h *handler) OnSecretChange(key string, secret *corev1.Secret) (*corev1.Secret, error) {
```
brandond, line 67: This needs to be reverted. This will list (without cache) all services in the secret's namespace, any time any secret in the cluster changes.

This should be reimplemented with an index on the service's secret annotation (so that services that reference a secret can be looked up from the index), and a RelatedResource handler on secrets to trigger index lookup -> change to service.

rohitsuse: Thank you @brandond I will re-work on this.

PR #613, opened 2026-01-19
[main] Make need a cert controllers trigger based on secret changes

rohitsakala:
Issue : https://github.com/rancher/rancher/issues/52097

### Summary 
- In short, whenever secret changes, the service related to it is enqueued in `needacert` package. 
- Added an indexer for sercert to service and then used `relatedresource` to enqueue `service` when `secret` changes.
- Inspired from https://github.com/k3s-io/helm-controller/pull/268/changes/dd15ed6b0019ddf3ebae2ea8b554d4146d042ac9

### DIFF
--- a/pkg/needacert/needacert.go
+++ b/pkg/needacert/needacert.go
@@ -389,6 +389,7 @@ func (h *handler) scheduleNextCertCheck(obj metav1.Object, secret *corev1.Secret
 
 	renewBefore := 24 * 60 * time.Hour
 	nextCheck := time.Until(cert.NotAfter.Add(-renewBefore))
+
 	if nextCheck < time.Minute {
 		nextCheck = time.Minute
 	}
--- a/pkg/needacert/needacert_test.go
+++ b/pkg/needacert/needacert_test.go
@@ -3,6 +3,7 @@ package needacert
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


joshmeranda (approved):

pkg/needacert/needacert.go
```diff
@@ -290,34 +326,42 @@ func (h *handler) generateSecret(service *corev1.Service) (*corev1.Secret, error
...
+		return nil, updateErr
 	} else if updated != nil {
-		return h.secrets.Update(updated)
+		secret, err = h.secrets.Update(updated)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err := h.scheduleNextCertCheck(service, secret); err != nil {
```
tomleb, line 357: nit: scheduleNextCertCheck calls parseCert but we already have the cert parsed here so we could just use it

pkg/needacert/needacert_test.go
```diff
@@ -798,32 +842,23 @@ func TestHandler_Race_Stress(t *testing.T) {
...
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "badsecret",
-			Namespace: "ns",
-		},
-		Type: corev1.SecretTypeTLS,
-		Data: map[string][]byte{
-			corev1.TLSCertKey:       []byte("-----BEGIN CERTIFICATE-----\nMIIB fake cert\n-----END CERTIFICATE-----"),
-			corev1.TLSPrivateKeyKey: []byte("not-a-key"),
+			Name: "badsecret",
```
tomleb: just curious why the "corrupted" data Secret was removed and replaced by just an empty secret

tomleb (approved):
let's see what Brad thinks but otherwise looks fine to me

pkg/needacert/needacert.go
rohitsuse: There is a possibility that the cert/secret can be updated in the `updateSecret` function above. So I recalculated parseCert.

pkg/needacert/needacert_test.go
rohitsuse: True. Fixed it.

PR #615, opened 2026-01-21
Bump to k8s 1.35

crobby:
Issue:  https://github.com/rancher/rancher/issues/53434


pkg/generated/controllers/admissionregistration.k8s.io/factory.go
brandond, line 1: whats with the mode change on all these empty files? making them executable seems wrong.

crobby: Not sure yet...keeping in draft until I sort that out

brandond: im also curious why these even exist lol

crobby: Files aren't actually empty.  I think they appeared empty in the diff because the only diff was the permissions.  Now that is resolved.

ericpromislow (commented):
The `lint.yaml` here is still using golangci-lint version 1.57 but steve has moved to 2.7.1 -- if you do this as well here you'll have to migrate `.golangci.json` to the new format.

Seeing how we're now on go 1.25 I would recommend doing that. There might be some pain involved because by default the new version complains about some lint violations that the old version let through.

brandond:
If you want to break out the golangci-lint , I'd do the bump in another PR that gets merged first. I had to do that for K3s due to the huge list of things that it wanted to fix:
* https://github.com/k3s-io/k3s/pull/13343

Probably worth taking the time to do that, as a contributing factor to that giant K3s PR was our disabling CI lint checks a while back in order to get a release out, without opening an issue to track re-enabling them.

crobby:
> If you want to break out the golangci-lint , I'd do the bump in another PR that gets merged first. I had to do that for K3s due to the huge list of things that it wanted to fix:
> 
> * [Update golangci-lint and re-enable CI linting step k3s-io/k3s#13343](https://github.com/k3s-io/k3s/pull/13343)
> 
> Probably worth taking the time to do that, as a contributing factor to that giant K3s PR was our disabling CI lint checks a while back in order to get a release out, without opening an issue to track re-enabling them.

I will spin-up another PR to do the golangci-lint bump.

PR #616, opened 2026-01-22
[main] bumped k8s deps to v0.35.0

vardhaman22:
https://github.com/rancher/rancher/issues/52957 and https://github.com/rancher/rancher/issues/53434
- bumped k8s dependencies to 0.35.0
- also added openapi model package generation

### DIFF
--- a/pkg/controller-gen/args/args.go
+++ b/pkg/controller-gen/args/args.go
@@ -15,8 +15,6 @@ type CustomArgs struct {
 	// BoilerplateContent is the actual boilerplate content that has been
 	// read. It will be added to every generated files.
 	BoilerplateContent []byte
-	// Open API model package name
-	OpenAPIModelPackageName string
 }
 
 type Options struct {


go.mod
```diff
@@ -7,22 +7,22 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/moby/locker v1.0.1
 	github.com/pkg/errors v0.9.1
-	github.com/rancher/lasso v0.2.5
+	github.com/rancher/lasso v0.2.6-rc.1
```
rohitsuse: My bad this should be [v0.2.6](https://github.com/rancher/lasso/releases/tag/v0.2.6). We don't generally rc lasso. I just found that.

vardhaman22: updated

pkg/controller-gen/main.go
```diff
@@ -325,16 +333,39 @@ func generateOpenAPI(groups map[string]bool, customArgs *cgargs.CustomArgs) erro
...
 	}
 
+	boilerplate, err := gengo.GoBoilerplate(openAPIArgs.GoHeaderFile, gengo.StdBuildTag, gengo.StdGeneratedBy)
+	if err != nil {
+		return err
+	}
+
 	getTargets := func(context *generator.Context) []generator.Target {
-		return oa.GetTargets(context, openAPIArgs)
+		return oa.GetOpenAPITargets(context, openAPIArgs, []byte(boilerplate))
```
rohitsuse:
```suggestion
		return oa.GetOpenAPITargets(context, openAPIArgs, boilerplate)
```

pkg/controller-gen/main.go
```diff
@@ -325,16 +333,39 @@ func generateOpenAPI(groups map[string]bool, customArgs *cgargs.CustomArgs) erro
...
+	); err != nil {
+		return err
+	}
+
+	if len(inputModelDirs) <= 0 {
+		return nil
+	}
+
+	getModelNameTargets := func(context *generator.Context) []generator.Target {
+		return oa.GetModelNameTargets(context, openAPIArgs, []byte(boilerplate))
```
rohitsuse:
```suggestion
		return oa.GetModelNameTargets(context, openAPIArgs, boilerplate)
```

pkg/controller-gen/args/args.go
```diff
@@ -15,6 +15,8 @@ type CustomArgs struct {
 	// BoilerplateContent is the actual boilerplate content that has been
 	// read. It will be added to every generated files.
 	BoilerplateContent []byte
+	// Open API model package name
+	OpenAPIModelPackageName string
```
rohitsuse: Where is this field being used ?

pkg/controller-gen/main.go
vardhaman22: done

pkg/controller-gen/main.go
vardhaman22: done

pkg/controller-gen/args/args.go
vardhaman22: it is used to add the open api model package comment so that the model file is generated.
https://github.com/rancher/wrangler/pull/616/changes#diff-a6bceab4ef640912a361c07d9bfab8a4772103b47f118d518fd414504c2eb213

vardhaman22: removed it...
it was added to the groups and i forgot to remove it form the custom args.

PR #620, opened 2026-02-09
Use watchlist with summary informer only if client supports it

tomleb:
# Issue https://github.com/rancher/rancher/issues/53854

When using a `cache.ListWatch`, Kubernetes assumes it will support the WatchList feature. However, that is not always the case since the client miht not support the new WatchList semantic.

The client-go library now includes a [function](https://github.com/kubernetes/client-go/blob/9bcb69436287b966d0c5c195efef00aed921fb1b/tools/cache/listwatch.go#L157) and an [interface](https://github.com/kubernetes/client-go/blob/9bcb69436287b966d0c5c195efef00aed921fb1b/util/watchlist/watch_list.go#L84-L86) that allows a client to opt-out of the WatchList feature if it doesn't support it.

We need this in Rancher to workaround a bug during an upgrade path. It's still useful imo for wrangler anyway.

Here's the accompanying PR that makes use of this: https://github.com/rancher/steve/pull/1034/.

### DIFF
--- a/pkg/summary/informer/watchlist.go
+++ b/pkg/summary/informer/watchlist.go
@@ -1,0 +1,54 @@
+package informer
+
+import (
+	"k8s.io/client-go/tools/cache"
+)
+
+// Copied from Kubernetes' source, only available in k8s 1.35
+
+type listWatcherWithWatchListSemanticsWrapper struct {
+	*cache.ListWatch
+
+	// unsupportedWatchListSemantics indicates whether a client explicitly does NOT support
+	// WatchList semantics.
+	//
+	// Over the years, unit tests in kube have been written in many different ways.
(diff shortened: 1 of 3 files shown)


pkg/summary/informer/watchlist.go
```diff
@@ -0,0 +1,54 @@
+package informer
+
+import (
+	"k8s.io/client-go/tools/cache"
+)
+
+// Copied from Kubernetes' source, only available in k8s 1.35
```
tomleb, line 7: I did it this way to still allow us using wrangler `main` with k8s 1.34. But hey maybe we'd rather just branch (as we branched already) and prevent k8s 1.34 from using this..

rohitsakala (approved):

PR #621, opened 2026-02-12
[release/v3.3] Revert 1.35 and 1.25

joshmeranda:
Reverting bump to k8s 1.35 and go 1.24


joshmeranda:
@crobby is 100% correct oops!

PR #623, opened 2026-02-13
Add k8s bump validation

tomleb:
# Issue https://github.com/rancher/rancher/issues/53766

We recently forgot to bump the minor version of wrangler after a Kubernetes minor bump. This needs to be done according to our ADR, it lets us to patch fixes for older versions IF absolutely needed.

We don't do Kubernetes minor bump often so it's easy to forget. This GHA workflow acts as a "forced" reminder. Essentially, if the CI detects that the k8s minor version changes by a PR, we'll require a manual acknowledgement from frameworks: someone will have to write `/ack-k8s-minor` in the comment of the PR.

I'll mark the job as "Required" so that we're sure it MUST succeed to be able to merge.

You can find an example here: https://github.com/tomleb/rancher-wrangler/pull/4.

### DIFF
--- a/.github/workflows/verify-k8s-bump.yaml
+++ b/.github/workflows/verify-k8s-bump.yaml
@@ -1,0 +1,200 @@
+name: Kubernetes Version Check
+
+# Note on Status Checks:
+# This workflow manages a custom status check named "k8s-version-ack".
+# In GitHub Branch Protection settings, you should specifically require "k8s-version-ack".
+#
+# DO NOT require the job named "Kubernetes Version Check / check-k8s-version", as 
+# it is designed to always pass (green) even when an acknowledgement is missing, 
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-k8s-bump.yaml
```diff
@@ -0,0 +1,200 @@
...
+              return;
+            }
+
+            const comments = await github.rest.issues.listComments({
+              owner,
+              repo,
+              issue_number: prInfo.number,
+            });
+
+            const authorizedRoles = ['admin', 'maintain', 'write'];
```
tomleb, line 121: Currently, only frameworks and architects and jono have write access.

Having write access gives us the ability to create tags, so in that sense, it makes sense that if you're able to create releases, you should be able to ackowledge that a new minor is required.

PR #624, opened 2026-02-13
Implement DeleteCollection.

bigkevmcd:
This adds support for DeleteCollection which allows for deleting resources using a ListOptions with Selector to select multiple resources in a namespace.

See https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/core/v1#ConfigMapInterface and https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/core/v1#SecretInterface for examples.

### DIFF
--- a/pkg/crd/mockCRDClient_test.go
+++ b/pkg/crd/mockCRDClient_test.go
@@ -25,6 +25,7 @@ import (
 type MockCustomResourceDefinitionInterface struct {
 	ctrl     *gomock.Controller
 	recorder *MockCustomResourceDefinitionInterfaceMockRecorder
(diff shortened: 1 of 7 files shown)


bigkevmcd:
Reviewing this code is what triggered me to add support...

https://github.com/rancher/rancher/pull/52942/changes#diff-91800a6701db50da2e7bc6fba022e8dda72baee53753c2989b509a8a1b42097cR18-R29

That would be reduced to a single-call, that shouldn't hit race-conditions :-)

bigkevmcd:
I'm not familiar with the processes for wrangler merges, is this ok to merge?

PR #628, opened 2026-03-03
Update packages to remove dep on archived `github.com/pkg/errors`

brandond:
This package has been archived since 2021 and ecosystem is finally moving off it. We should as well.

This removes all direct use of the package. On the master branch, all deps have also moved off it.

errors.Wrap/Wrapf/WithMessage can all be replaced by stdlib fmt.Errorf with %w to wrap the error.

This may also bring some performance increases as Wrap/Wrapf capture a full stack trace from the caller which we do not (to my knowledge) ever actually use.

* Issue: https://github.com/rancher/rancher/issues/54013


brandond:
@crobby created an issue

PR #647, opened 2026-04-07
[v3.3] pin mockgen to commit SHA instead of mutable tag

crobby:
Pin go install for mockgen from the v0.5.0 tag to its underlying commit SHA

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,7 @@ jobs:
       # https://github.com/actions/checkout/releases/tag/v4.1.1
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
     - name : Install mockgen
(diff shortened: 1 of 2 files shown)


brandond (changes requested):
This is unnecessary. While the git tag may be mutable, the go sumdb will complain if the hash it points at is changed. This is the entire point of go.sum and the go module proxy.

See: https://go.dev/ref/mod#authenticating

tl;dr we do not need to switch from referencing golang module versions to hashes

crobby:
> This is unnecessary. While the git tag may be mutable, the go sumdb will complain if the hash it points at is changed. This is the entire point of go.sum and the go module proxy.
> 
> See: https://go.dev/ref/mod#authenticating
> 
> tl;dr we do not need to switch from referencing golang module versions to hashes

Ah, yes, you are correct.  I just double-checked our guide and this should be unnecessary.  Thanks.

Issue #652, opened 2026-04-28
Goroutine leak in summaryResourceClient.Watch due to missing context cancellation


vvlisn:
## Problem

The `Watch` function in `pkg/summary/client/simple.go` creates a goroutine that forwards events from the underlying watch response to a local unbuffered channel. This goroutine has no context cancellation mechanism, causing it to leak when the consumer stops reading from the channel.

### Root Cause

In [`pkg/summary/client/simple.go:108-117`](https://github.com/rancher/wrangler/blob/main/pkg/summary/client/simple.go#L108-L117):

```go
func (c *summaryResourceClient) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
    eventChan := make(chan watch.Event)  // unbuffered channel
    // ...
    go func() {
        defer close(eventChan)
        for event := range resp.ResultChan() {
            if _, ok := event.Object.(*metav1.Status); !ok {
                event.Object = summary.SummarizedWithOptions(event.Object, generateSummarizeOpts(c.options.Schema))
            }
            eventChan <- event  // blocks forever when consumer stops reading
        }
    }()
    // ...
}
```

The goroutine blocks on `eventChan <- event` when the consumer (e.g., an informer) stops reading from the channel. Since there is no `select` with `ctx.Done()`, the goroutine can never exit and leaks permanently.

### Evidence from Production

Analyzed a cluster-agent pprof profile after ~19 days of runtime:

- **66 goroutines** permanently stuck in `chan send` state at `simple.go:115`
- Longest blocked goroutine: **27,528 minutes (~19 days)** — nearly the entire agent lifetime
- Each watch reconnection leaks one goroutine, accumulating over time
- Total goroutines: 5,103 (with 206 created by `summaryResourceClient.Watch`)

Goroutine stack trace:

```
goroutine 58656017 [chan send, 12513 minutes]:
github.com/rancher/wrangler/v3/pkg/summary/client.(*summaryResourceClient).Watch.func1()
        .../pkg/summary/client/simple.go:115 +0xc6
created by ...(*summaryResourceClient).Watch
        .../pkg/summary/client/simple.go:108 +0x265
```

### Impact

- Goroutines accumulate indefinitely, holding references to watch event objects
- Prevents GC from reclaiming memory associated with those events
- Contributes to continuous memory growth in cluster-agent (observed: 170.87MB heap after 19 days)
[...]

PR #661, opened 2026-05-01
Plug a goroutine leak.

ericpromislow:
Applied @vvlisn's suggested fix from [652](https://github.com/rancher/wrangler/issues/652)

### DIFF
--- a/pkg/summary/client/simple.go
+++ b/pkg/summary/client/simple.go
@@ -112,7 +112,11 @@ func (c *summaryResourceClient) Watch(ctx context.Context, opts metav1.ListOptio
 			if _, ok := event.Object.(*metav1.Status); !ok {
 				event.Object = summary.SummarizedWithOptions(event.Object, generateSummarizeOpts(c.options.Schema))
 			}
-			eventChan <- event
+			select {
+			case eventChan <- event:
+			case <-ctx.Done():
+				return
+			}
 		}
 	}()
 
--- a/pkg/summary/client/simple.go
+++ b/pkg/summary/client/simple.go
@@ -94,6 +94,7 @@ func (c *summaryResourceClient) Watch(ctx context.Context, opts metav1.ListOptio
 		err  error
 	)
 
+	ctx, cancel := context.WithCancel(ctx)
 	eventChan := make(chan watch.Event)
 
 	if c.namespace == "" {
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
Question: are we confident that readers are going away because the context is cancelled? Are there any other conditions where the reader may stop reading, but the context continues to be valid?

pkg/summary/client/simple.go
brandond, line 111: In looking at https://pkg.go.dev/k8s.io/apimachinery/pkg/watch#Interface - as the consumer of this watch, we are responsible for stopping it.

>	// Because some implementations may create channels when constructed, Stop
>	// must always be called, even if the consumer has not yet called
>	// ResultChan().

I think this means we should probably also `defer resp.Stop()` here as well?

aruiz14: No, that shouldn't happen, as the returned `watcher` struct only overrides `ResultChan()` from the embedded `watch.Interface`. The caller to `summaryResourceClient.Watch()` is responsible for calling `Stop()`, which actually acts on the inner `resp`.

pkg/summary/client/simple.go
```diff
@@ -112,7 +112,11 @@ func (c *summaryResourceClient) Watch(ctx context.Context, opts metav1.ListOptio
 			if _, ok := event.Object.(*metav1.Status); !ok {
 				event.Object = summary.SummarizedWithOptions(event.Object, generateSummarizeOpts(c.options.Schema))
 			}
-			eventChan <- event
+			select {
+			case eventChan <- event:
+			case <-ctx.Done():
```
aruiz14, line 120: While this may fix the issue if the caller context is correctly canceled, I believe it should have its own context that gets canceled on `Stop()`. I believe we should create a derived context with `WithCancel`, then make `Stop()` cancel it, in addition to the embedded `resp.Stop()`

aruiz14 (changes requested):

ericpromislow:
@aruiz14 Would you be able to take over from me? I only see how it's used from `steve`, and then I'm only really familar with running `steve` standalone, in which case it runs forever until it's killed, so nothing needs to be gracefully closed.

pkg/summary/client/simple.go
brandond: Yeah... I agree. Relying on the embedded resp's `Stop()` to be called and handle everything seems fragile. `watcher` should implement its own Stop that ensures the context is cancelled to stop the goroutine.

brandond:
@ericpromislow @aruiz14 how about that?

brandond (approved):

aruiz14 (approved):

prachidamle (approved):

PR #668, opened 2026-05-08
misc(summaryInformer): make ListWatcher context-aware

aruiz14:
`ListFunc` and `WatchFunc` are marked as deprecated (`context.TODO()` was already a hint), in favor of context-aware equivalents.

### DIFF
--- a/pkg/summary/informer/informer.go
+++ b/pkg/summary/informer/informer.go
@@ -126,17 +126,17 @@ func NewFilteredSummaryInformerWithOptions(
 	tweakListOptions TweakListOptionsFunc,
 ) informers.GenericInformer {
 	lw := &cache.ListWatch{
-		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+		ListWithContextFunc: func(ctx context.Context, options metav1.ListOptions) (runtime.Object, error) {
 			if tweakListOptions != nil {
 				tweakListOptions(&options)
 			}
-			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).List(context.TODO(), options)
+			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).List(ctx, options)
 		},
-		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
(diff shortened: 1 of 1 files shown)


brandond (approved):
Wangler in general makes excessive use of `context.TODO()`; when context was introduced to the various client-go calls we just punted instead of making the hard change to plumb context through as we should have.

brandond (changes requested):
While you're here, why not do the other ListWatch a couple lines above in `NewFilteredSummaryInformerWithOptions`?

tomleb:
> Wangler in general makes excessive use of `context.TODO()`; when context was introduced to the various client-go calls we just punted instead of making the hard change to plumb context through as we should have.

@brandond  Not 100% sure, but I think the reason for this is because wrangler was semver (now semver-ish) so using new context-aware methods / fields would break old codebases if you were using pinning to pin k8s to older version.

We've talked a few times now about making wrangler making wrangler basically just follow k8s in terms of Go and k8s libraries.. I still haven't "done" that yet, not sure if it requires a RFD of some kind.. But I think we should just do it because we keep coming back to the same annoying and confusing rules of "well, does this break an older version of rancher / k3s if we're pinning old version.

brandond:
Yeah, I guess my point is we (Darren) should have made a breaking change to wrangler when upstream client-go made their breaking change. Instead we had Wrangler just inject TODO everywhere and now it's years later and there is MORE downstream code to fix, not less.

I'm just whinging.

pkg/summary/informer/informer.go
brandond, line 141: add contexts here too

brandond (changes requested):

pkg/summary/informer/informer.go
aruiz14: good catch, thanks!

brandond (approved):

rancher-max:
This provides a partial fix for https://github.com/rancher/rancher/issues/55624. Can we aim to get this into October releases? (I know Rancher hits September code freeze today, so it doesn't make sense to put it in there)

crobby (approved):

PR #679, opened 2026-06-19
fix(summary/informer): use informer context in List and Watch calls

### LINKED ISSUE #55624: [BUG]  CRD with broken conversion webhook makes entire cluster unresponsive in Rancher UI/API
**Rancher Server Setup**
- Rancher version: v2.14.1
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):  RKE2 v1.35.4+rke2r1, single all-role node, custom cluster
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.35.4+rke2r1
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom 

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin

**Describe the bug**

When a CRD that includes a conversion webhook configuration is left over in a downstream cluster after its provider has been uninstalled, the webhook service no longer exists. Any attempt by Rancher/cattle-cluster-agent to list or watch objects of that CRD kind causes the Kubernetes API server to return a webhook timeout error.

The Rancher-side bug is that `steve` holds the cluster cache write lock `(h.Lock() / defer h.Unlock())` for the entire duration of OnSchemas, which includes a `WaitForCacheSync` call with a 15-minute timeout per broken GVK. Because every concurrent `Get/List` call in the cluster cache blocks behind that same mutex, one broken CRD freezes all unrelated API reads (Pods, Namespaces, Deployments, etc.) for up to 15 minutes.

A secondary bug in `wrangler` compounds this: the summary informer uses `context.TODO()` in its `ListFunc/WatchFunc` closures instead of the reflector's run context, so cancelling a broken watcher does not interrupt its in-flight API request — the request keeps waiting on the full Kubernetes webhook timeout regardless.

The combined effect is that loading any webpage or API related to the affected downstream cluster takes 10–20 minutes but eventually returns the correct content.

**To Reproduce**

1. Provision a Rancher v2.14.1 downstream cluster (single all-role node, RKE2 v1.35.4+rke2r1 custom cluster).
2. Apply the `buckets.s3.aws.upbound.io.yaml CRD` into the downstream cluster
3. Create the `bucket` CR into the downstream cluster.
4. Delete the webhook service (or simply do not install the provider), leaving the CRD with a broken webhook reference.
5. Observe `cattle-cluster-agent` logs — the `"Failed to watch"` error begins repeating immediately.
6. Attempt to open the affected cluster in the Rancher UI or call any cluster API endpoint (e.g., list Pods, Namespaces).
 
 **Result**

1. `cattle-cluster-agent` logs `"Failed to watch … conversion webhook … service not found"` repeatedly with exponential back-off.
2. All Rancher UI pages and API calls for the affected cluster are slow (10–20 minutes) before returning a result.
3. `kubectl get bucket` returns the webhook error immediately.
4. Unrelated resources (Pods, Namespaces, Deployments) are also inaccessible during the timeout window despite not being related to the broken CRD.

**Expected Result**

The broken CRD/webhook is a cluster configuration problem and Rancher cannot fix it automatically. However, Rancher should remain fully responsive for all unrelated cluster resources. The broken watcher for the Bucket CRD should fail in isolation without blocking reads for Pods, Namespaces, Deployments, or any other GVK.

**Screenshots**

**Additional context**

khushalchandak17:
Partially Fixes rancher/rancher#55624

`NewFilteredSummaryInformer` and `NewFilteredSummaryInformerWithOptions` used
`context.TODO()` in their `ListFunc`/`WatchFunc` closures. When a watcher is
cancelled, the in-flight API request had no way to receive that cancellation
and would keep waiting on the full Kubernetes webhook timeout.

Replace `ListFunc`/`WatchFunc` with `ListWithContextFunc`/`WatchFuncWithContext`
and pass through the informer's run context so cancellation propagates correctly.

### DIFF
--- a/pkg/summary/informer/informer.go
+++ b/pkg/summary/informer/informer.go
@@ -126,17 +126,17 @@ func NewFilteredSummaryInformerWithOptions(
 	tweakListOptions TweakListOptionsFunc,
 ) informers.GenericInformer {
 	lw := &cache.ListWatch{
-		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+		ListWithContextFunc: func(ctx context.Context, options metav1.ListOptions) (runtime.Object, error) {
 			if tweakListOptions != nil {
 				tweakListOptions(&options)
 			}
-			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).List(context.TODO(), options)
+			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).List(ctx, options)
 		},
-		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+		WatchFuncWithContext: func(ctx context.Context, options metav1.ListOptions) (watch.Interface, error) {
 			if tweakListOptions != nil {
 				tweakListOptions(&options)
 			}
-			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).Watch(context.TODO(), options)
+			return client.ResourceWithOptions(gvr, opts).Namespace(namespace).Watch(ctx, options)
 		},
 	}
 	return &summaryInformer{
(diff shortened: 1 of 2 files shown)


brandond:
* Duplicate of https://github.com/rancher/wrangler/pull/668

Any reason not to move forward with the existing PR?

khushalchandak17:
Thanks, I missed https://github.com/rancher/wrangler/pull/668 when preparing this fix. The production change is indeed the same.  There is no reason not to proceed with the existing PR, this PR can be closed as a duplicate.

PR #687, opened 2026-07-15
Make CABundle CA-chaining optional in needacert

crobby:
Issue: https://github.com/rancher/rancher/issues/56116
Issue:  https://github.com/rancher/turtles/issues/2452

Support service annotation to enable/disable "ca-only" mode vs default of chaining all certs together.

### DIFF
--- a/pkg/needacert/needacert.go
+++ b/pkg/needacert/needacert.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


tomleb (approved):
LGTM

crobby:
It seems like this change does address the needs for the turtles team.  (note to self)...They're requesting a backport for 2.15.1 as well.

pkg/needacert/needacert.go
```diff
@@ -107,6 +126,41 @@ type handler struct {
...
+	}
+	if len(certs) == 0 {
+		return nil, fmt.Errorf("no certificates found in chain")
+	}
+
+	caCert := certs[len(certs)-1]
+	var buf bytes.Buffer
+	if err := pem.Encode(&buf, &pem.Block{Type: cert.CertificateBlockType, Bytes: caCert.Raw}); err != nil {
+		return nil, fmt.Errorf("failed to encode CA certificate: %w", err)
+	}
```
brandond, line 165: nit/potential optimization: if `len(certs) == 1` then you can just return `fullChainPEM` as-is without re-encoding.

brandond (approved):

PR #693, opened 2026-07-22
test: add unit tests for Export and ToBytes functions

rohitsuse:
Adds missing unit tests for two functions.

### DIFF
--- a/pkg/yaml/export_test.go
+++ b/pkg/yaml/export_test.go
@@ -1,0 +1,50 @@
+package yaml
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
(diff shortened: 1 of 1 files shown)


brandond:
Is there any specific behavior from the old library that we are looking to test? This just seems to be confirming that the yaml library successfully round-trips things, without validating any specific behavior.

In general I don't see a lot value in tests that just validate 3rd party library functionality, without some note as to what specific thing we're looking for or are concerned about breaking in the future.

rohitsuse:
> Is there any specific behavior from the old library that we are looking to test? This just seems to be confirming that the yaml library successfully round-trips things, without validating any specific behavior.
> 
> In general I don't see a lot value in tests that just validate 3rd party library functionality, without some note as to what specific thing we're looking for or are concerned about breaking in the future.

 You're right. I've refocused the tests on the functions' own behavior rather than round-tripping the library.  Marshalling is exercised as a side effect, but the assertions are on wrangler's logic, not the YAML library.

brandond (approved):

PR #696, opened 2026-07-30
Support Rancher operation CRDs

jiaqiluo:
**Issue**:
- https://github.com/rancher/rancher/issues/56075

**Problem**:

Operation conditions appear incorrectly in the UI, showing red with the text `Pending`, even though the operation has moved to a different state. 
This is due to a quirk within Wrangler's summarization.

**Fix**:

Adds first-class summary/transitioning support for Rancher `operation.cattle.io` CRDs, `EncryptionKeyRotation`, `ETCDSnapshotRestore`, and `ETCDSnapshotSave`, so their lifecycle conditions map to the correct Wrangler Summary state/flags rather than relying on the generic condition tables.

**Dev Validation**:

The changes were validated on a customized Rancher image where the Wrangler module was replaced with this branch.  I confirmed that the UI showed the color and message accurately for those three CRDs. 

ex2

eample

ex4

Screenshot 2026-07-30 at 12 17 45 PM

### DIFF
--- a/pkg/summary/operations.go
+++ b/pkg/summary/operations.go
@@ -12,10 +12,11 @@ var operationKinds = map[string]bool{
 	"EncryptionKeyRotation": true,
 	"ETCDSnapshotRestore":   true,
 	"ETCDSnapshotSave":      true,
+	"CertificateRotation":   true,
 }
 
 // isOperation returns true if the object is a Rancher operation
-// (operation.cattle.io/*/{EncryptionKeyRotation,ETCDSnapshotRestore,ETCDSnapshotSave}).
+// (operation.cattle.io/*/{EncryptionKeyRotation,ETCDSnapshotRestore,ETCDSnapshotSave,CertificateRotation}).
 func isOperation(obj data.Object) bool {
 	return strings.HasPrefix(obj.String("apiVersion"), "operation.cattle.io/") &&
 		operationKinds[obj.String("kind")]
--- a/pkg/summary/operations.go
+++ b/pkg/summary/operations.go
@@ -1,0 +1,85 @@
+package summary
+
+import (
+	"strings"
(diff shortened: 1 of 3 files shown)


pkg/summary/operations.go
```diff
@@ -0,0 +1,85 @@
...
+			summary.State = state.state
+			if state.err {
+				summary.Error = true
+			}
+			if state.transitioning {
+				summary.Transitioning = true
+			}
+			if msg := c.Message(); msg != "" {
+				summary.Message = append(summary.Message, msg)
+			}
```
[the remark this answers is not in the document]
jiaqiluo: It's not an issue because those 3 CRDs do not have the `.status.display` or `.status.summary` fields that the `checkStatusSummary` looks for.

pkg/summary/operations.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/rancher/wrangler/v3/pkg/data"
+)
+
+// operationKinds are the operation.cattle.io kinds that embed the shared
+// OperationStatus contract.
+var operationKinds = map[string]bool{
+	"EncryptionKeyRotation": true,
+	"ETCDSnapshotRestore":   true,
+	"ETCDSnapshotSave":      true,
```
jferrazbr, line 14: could you also add `CertificateRotation` here? I will open its [PR](https://github.com/rancher/rancher/pull/56333) very soon, and including it now would avoid a follow-up PR to add it

jiaqiluo: Cool. CertificateRotation was added!

jferrazbr (approved):

jakefhyde (approved):

tomleb (approved):

PR #698, opened 2026-08-13
Add AgentRestart field to schemas.Field for KDM agent-restart metadata

pratikjagrut:
Part of https://github.com/rancher/rancher/issues/55326
## What this does

  Adds an optional `AgentRestart` boolean field to `schemas.Field` (`pkg/schemas/types.go`).

  This lets KDM `serverArgs`/`agentArgs` entries carry an `agentRestart` flag indicating that
  toggling the argument requires affected agents to restart to pick up the change.

  ## Why

  Some server-only flags (e.g. `disable-kube-proxy`) affect agent behavior but are stripped
  from worker config, so workers never restart when they change. Making this data-driven via
  KDM lets Rancher decide which flags force an agent restart without hardcoding the list.

  ## Related PRs

  - KDM: https://github.com/rancher/kontainer-driver-metadata/pull/2156
  - Rancher: https://github.com/rancher/rancher/pull/56603

### DIFF
--- a/pkg/schemas/types.go
+++ b/pkg/schemas/types.go
@@ -114,6 +114,7 @@ type Field struct {
 	InvalidChars string      `json:"invalidChars,omitempty"`
 	Description  string      `json:"description,omitempty"`
 	CodeName     string      `json:"-"`
+	AgentRestart bool        `json:"agentRestart,omitempty"`
 }
 
 type Action struct {


brandond (approved):

crobby (approved):

snasovich (changes requested):
Adding provisioning/KDM-specific flag to what looks like _very_ base type `Field` doesn't look right to me.
Copilot/Claude Opus 4.8 agrees and provides more context / suggestions for better approaches - with primary one avoiding need for any Wrangler-side changes.

---
Your instinct is right — extending `schemas.Field` is a heavier hammer than this feature needs. Here's the analysis.

## What's actually happening

In KDM, each `serverArgs`/`agentArgs` entry is modeled as a `schemas.Field`. The PR adds a new field to that struct:

```go name=pkg/schemas/types.go url=https://github.com/rancher/wrangler/blob/fd95fdf558678d4ae0351418f2cf3abb55403e45/pkg/schemas/types.go#L100-L118
type Field struct {
	Type         string      `json:"type,omitempty"`
	...
	CodeName     string      `json:"-"`
	AgentRestart bool        `json:"agentRestart,omitempty"`
}
```

The problem you're sensing is real: `schemas.Field` is a **general-purpose Norman schema primitive**. It's used far beyond KDM args — dynamic schemas for node drivers, KEv2 cloud-credential schemas, embedded credential configs, and every generated Norman resource field. For example:

- `pkg/data/management/kev2_creds_schema.go` builds cloud-credential schemas from `v32.Field`
- `pkg/controllers/management/drivers/nodedriver/machine_driver.go` builds dynamic node-driver schemas from `Field`
- All the `zz_generated_*` Norman clients

Adding `AgentRestart` puts a KDM/CAPR-provisioning-specific concept onto a type that describes API resource fields generically. It leaks a domain concept into a widely-shared struct, and it becomes part of the serialized JSON contract (`agentRestart`) for every schema everywhere, not just KDM.

## Why the change targets `Field` anyway

The consuming side already reads `schemas.Field` directly from KDM:

```go name=pkg/capr/planner/filter.go url=https://github.com/rancher/rancher/blob/59bd052e44553e77e5c570b84ed4a24dfa7b44ab/pkg/capr/planner/filter.go#L45-L74
field, fieldFound := release.AgentArgs[k]
if !fieldFound && isServer {
	field, fieldFound = release.ServerArgs[k]
}
...
if field.Type == "boolean" {
```

So `release.ServerArgs`/`AgentArgs` are `map[string]schemas.Field`, and the path of least resistance is to bolt the flag onto `Field`. That's convenience, not correctness.

## Narrower alternatives worth raising on the PR

1. **A dedicated KDM-side type (preferred).** The `model.Release`'s `ServerArgs`/`AgentArgs` are defined in KDM (`rancher/kontainer-driver-metadata`), not wrangler. If those maps used a KDM-owned struct (e.g. `model.Arg` that embeds/mirrors the needed field metadata plus `AgentRestart`), the new concept lives entirely in the provisioning domain and never touches wrangler's shared `Field`. This is the cleanest separation and avoids a wrangler release/bump altogether.

2. **Piggyback on existing extensibility instead of a new typed field.** `Schema` already has a generic `Attributes map[string]interface{}` for exactly this kind of side-channel metadata (see `SetHasObservedGeneration`/`HasObservedGeneration`). If it must stay in wrangler, an attribute-style approach keeps `Field`'s typed surface stable — though `Field` itself has no attributes map, which is another argument for option 1.

3. **If it truly must be a typed field on `Field`,** at minimum it should be justified as a generic schema capability (it isn't — "agent restart" is meaningless outside CAPR provisioning), which further supports moving it to KDM.

brandond:
yeah, I'm :-1: on this being a Field attribute. Find somewhere in the KDM data model for it.

pratikjagrut:
Hi @snasovich,
Thank you for the review. I’m trying to rewrite the solution based on your suggestion of having a dedicated KDM-side type.

I couldn’t find a `model.Release` type in the KDM repo. I believe you meant `rancher/channelsserver` repo, if I’m not mistaken.

https://github.com/rancher/channelserver/blob/master/pkg/model/config.go#L22

brandond:
Yes, the channel server defines the data model. Rancher, K3s, and RKE2 each have channel or release files that are consumed and published by channel server.

pratikjagrut:
Closing this PR in favour of https://github.com/rancher/channelserver/pull/45

PR #713, opened 2026-09-01
Add support for server-side apply (field management) in generated clients and fakes

brandond:
Wrangler's code-gen helpers do not currently enable support for field management (server-side apply) in clients. In order to do this, we need to generate applyconfiguration code, and reference it when generating the clients.
* Ref: https://github.com/kubernetes/kubernetes/pull/125560

Fake NewClientset requires OpenAPI types to function properly; without
OpenAPI type generation enabled only NewSimpleClientset is usable for
custom types.
* Ref: https://github.com/kubernetes/kubernetes/issues/126850

Server-side apply has been stable since 1.22 so we are somewhat behind on supporting it:
* https://kubernetes.io/docs/reference/using-api/server-side-apply/
* https://pkg.go.dev/k8s.io/client-go/applyconfigurations

### DIFF
--- a/pkg/controller-gen/generators/models_schema.go
+++ b/pkg/controller-gen/generators/models_schema.go
@@ -1,0 +1,86 @@
+package generators
+
+import (
+	"path"
+
+	"github.com/rancher/wrangler/v3/pkg/controller-gen/args"
+	"k8s.io/gengo/v2/generator"
+)
+
+func ModelsSchemaCmdGo(customArgs *args.CustomArgs) generator.Generator {
+	return &modelsSchemaCmdGo{
+		customArgs: customArgs,
+		GoGenerator: generator.GoGenerator{
+			OutputFilename: "main.go",
+			OptionalBody:   []byte(modelsSchemaCmdBody),
+		},
+	}
+}
+
+type modelsSchemaCmdGo struct {
+	generator.GoGenerator
+
+	customArgs *args.CustomArgs
+}
+
+func (m *modelsSchemaCmdGo) Imports(*generator.Context) []string {
+	return []string{
+		"encoding/json",
+		"fmt",
+		"os",
+		path.Join(m.customArgs.Package, "openapi"),
+		"k8s.io/kube-openapi/pkg/common",
+		"k8s.io/kube-openapi/pkg/validation/spec",
+	}
+}
+
+var modelsSchemaCmdBody = `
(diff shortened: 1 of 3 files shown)


tomleb (approved):
LGTM. I thought we were generating the clientset for core apis directly in wrangler, but it seems we aren't.

gehrkefc (approved):

joshmeranda (approved):

brandond:
We should probably add Apply/ApplyStatus to the controller's client as well but that will need to be done manually.
* Ref: https://github.com/rancher/wrangler/pull/624

brandond:
Turns out we need to actually generate the applyconfiguration package with `applyconfiguration-gen` too.

pkg/controller-gen/generators/models_schema.go
```diff
@@ -0,0 +1,86 @@
...
+		"fmt",
+		"os",
+		path.Join(m.customArgs.Package, "openapi"),
+		"k8s.io/kube-openapi/pkg/common",
+		"k8s.io/kube-openapi/pkg/validation/spec",
+	}
+}
+
+var modelsSchemaCmdBody = `
+// Outputs openAPI schema JSON containing the schema definitions in zz_generated.openapi.go.
```
brandond, line 38: cribbed from: https://github.com/kubernetes/kubernetes/blob/v1.36.4/pkg/generated/openapi/cmd/models-schema/main.go

pkg/controller-gen/main.go
```diff
@@ -234,6 +240,65 @@ func generateDeepcopy(groups map[string]bool, customArgs *cgargs.CustomArgs) err
...
+	)
+}
+
+func generateOpenAPISchemaFile(customArgs *cgargs.CustomArgs) (string, error) {
+	modelsSchemaGo := filepath.Join(customArgs.OutputBase, customArgs.Options.OutputPackage, "openapi", "cmd", "models-schema", "main.go")
+	if _, err := os.Stat(modelsSchemaGo); err != nil {
+		return "", errors.New("OpenAPI models-schema command not found; please enable GenerateOpenAPI for one or groups")
+	}
+
+	cmd := exec.Command("go", "run", modelsSchemaGo)
```
brandond, line 285: this is literally how upstream is doing it, except they do it from a shell script: https://github.com/kubernetes/kubernetes/blob/v1.36.4/staging/src/k8s.io/kube-aggregator/hack/update-codegen.sh#L51

PR #741, opened 2026-09-14
Add WithNullSafePatch for applying explicit nulls to CRDs

marytlf:
## Problem

`apply` cannot express an explicit `null` on any CRD.

`patch.GetMergeStyle` (`pkg/patch/style.go:59-60`) selects `types.MergePatchType` for
every GVK that is not registered in `k8s.io/client-go/kubernetes/scheme` — which is
every CRD. RFC 7386 defines a `null` member in a merge patch as an instruction to
**remove** the key, not as a value, so a caller that puts a literal `null` in the
desired object cannot get one onto the server.

The loss happens at patch generation, not on the server.
`jsonmergepatch.CreateThreeWayJSONMergePatch` builds the patch in two halves and runs
both through `keepOrDeleteNullInJsonPatch`: the add/change half strips nulls, and the
delete half keeps only nulls. An explicit null is therefore either discarded or turned
into a deletion. There is no error, event, or retry — the caller's intent just
disappears.

Downstream symptom: [rancher/rancher#56335](https://github.com/rancher/rancher/issues/56335).
Setting a Helm value to `null` in `spec.rkeConfig.chartValues` — the documented way to
remove a chart default — is silently dropped, so the generated `HelmChartConfig` never
removes the default.

## Solution

Add an opt-in, per-GVK patch mode that translates the merge patch `apply` already
computed into an equivalent RFC 6902 JSON Patch. RFC 6902 is the only patch format
Kubernetes accepts that distinguishes `{"op":"remove"}` from `{"op":"add","value":null}`.

### New API

```go
// pkg/apply/apply.go
WithNullSafePatch(gvks ...schema.GroupVersionKind) Apply
```

Implemented on `desiredSet` following the copy-on-write shape of the existing
`WithPatcher` / `WithReconciler`, backed by a `nullSafePatch` set on the struct.

### New: `pkg/patch/jsonpatch.go`

```go
func CreateJSONPatchFromMergePatch(mergePatch, modified, current []byte) ([]byte, error)
```

The ambiguity of a null is resolved against `modified`, the desired object the merge
patch was generated from:

| null in merge patch at path | in `modified` | in `current` | emitted |
|---|---|---|---|
| yes | explicitly `null` | — | `{"op":"add","value":null}` — **assignment** |
| yes | absent | present | `{"op":"remove"}` |
| yes | absent | absent | nothing (RFC 6902 `remove` on a missing path fails the whole patch) |

Everything the merge patch would have done is preserved. Supporting details:

- `Operation.Value` is a `*json.RawMessage`, so a literal null is distinguishable from
  an absent value — `remove` carries no `value` member, assigning null requires one.
- `add` rather than `replace`: on an object member RFC 6902 `add` is an upsert, correct
  whether or not the key exists, while `replace` fails on a missing target.
- Recursion only descends where `current` already holds an object, which guarantees the
  parent of every emitted path exists — the precondition RFC 6902 `add` requires.
- Where a subtree is not mergeable it is set wholesale from `modified`, not from the
  patch, so nulls it carries are assignments rather than stale removals.
- Arrays are opaque leaves, replaced wholesale — matches merge-patch semantics and keeps
  positional index ops, and their races, out of the result.
- Keys are walked in sorted order so the output is deterministic for tests and debug
  logs. Ordering is not needed for correctness: an op is emitted for a node or for its
  descendants, never both.

### Call site

One branch in `applyPatch` (`pkg/apply/desiredset_compare.go`), after the patch is
computed and before it is sent:

```go
if nullSafe && patchType == types.MergePatchType {
    _, strippedModified, strippedCurrent, err := stripIgnores(nil, modified, current, diffPatches)
    ...
    patch, err = patch2.CreateJSONPatchFromMergePatch(patch, strippedModified, strippedCurrent)
    patchType = types.JSONPatchType
    if string(patch) == "[]" { return false, nil }
}
```

The translation resolves against the ignore-stripped documents, the same ones `doPatch`
generated the merge patch from — otherwise a wholesale subtree would reinstate a field
an ignore patch had taken out.

### Three places that assumed the patch is a JSON object

1. `sanitizePatch` unmarshalled into a map, which errors on `[...]`. It now early-returns
   for array patches, via `dropObjectSetOps` when the objectset annotation must be
   stripped.
2. The empty-patch sentinel `"{}"` — also `"[]"` for this style, in `applyPatch` and in
   `createPlan` (`desiredset_process.go`). Missing it would issue a no-op `PATCH` on
   every reconcile.
3. `FakeApply` gets the passthrough needed to satisfy the interface.

## What is deliberately unchanged

`doPatch`, `GetPatchStyle`, `GetMergeStyle` and `patch.Apply` are untouched. A
`desiredSet` with no `nullSafePatch` entry for a GVK produces byte-identical patches to
today. This matters because `clients.Apply` is shared by every controller in Rancher —
the opt-in is per GVK, so enabling it for one kind leaves every other consumer
unaffected.

## Testing

Unit — `pkg/patch/jsonpatch_test.go`:

- an explicit null is assigned, not removed;
- a null for a key absent from `modified` is a removal;
- a removal of a key absent from `current` emits nothing;
- pointer tokens are escaped (RFC 6901 `~`→`~0`, `/`→`~1`, including `objectset.rio.cattle.io/applied`);
- arrays are replaced wholesale;
- a subtree missing from `current` is added wholesale with its nulls;
- a stale removal inside a new subtree is dropped;
- an object replacing a scalar is set wholesale;
- operations are emitted in a stable order;
- invalid input is rejected.

`pkg/apply/desiredset_compare_test.go`: `TestApplyPatchNullSafe` and
`TestSanitizePatchJSONPatch`.

Manual / integration — verified on a live cluster that the apiserver stores a null
written by `JSONPatchType` into an `x-kubernetes-preserve-unknown-fields` subtree rather
than pruning it. This is the one assumption no unit test can reach.

## Downstream

This PR alone fixes nothing; it adds a capability. The companion `rancher/rancher` PR
registers it for `RKEControlPlane`. Sequencing: merge here → tag → bump
`github.com/rancher/wrangler/v3` in `rancher/rancher` (currently `v3.7.1`) → merge the
Rancher PR. Each backport branch needs its own bump.

## Backport

New API surface only, no behaviour change for existing callers. Safe to backport to any
`v3.x` line that needs it.

### DIFF
--- a/pkg/apply/desiredset_compare.go
+++ b/pkg/apply/desiredset_compare.go
@@ -131,6 +131,13 @@ func emptyMaps(data map[string]interface{}, keys ...string) bool {
 }
 
 func sanitizePatch(patch []byte, removeObjectSetAnnotation bool) ([]byte, error) {
+	if patch2.IsJSONPatch(patch) {
+		if !removeObjectSetAnnotation {
+			return patch, nil
+		}
+		return dropObjectSetOps(patch)
+	}
+
 	mod := false
 	data := map[string]interface{}{}
 	err := json.Unmarshal(patch, &data)
@@ -185,7 +192,31 @@ func sanitizePatch(patch []byte, removeObjectSetAnnotation bool) ([]byte, error)
 	return json.Marshal(data)
 }
 
-func applyPatch(gvk schema.GroupVersionKind, reconciler Reconciler, patcher Patcher, debugID string, ignoreOriginal bool, oldObject, newObject runtime.Object, diffPatches [][]byte) (bool, error) {
+// objectSetOpPrefix is the JSON Pointer prefix of the operations that carry
+// apply's own bookkeeping annotations.
+var objectSetOpPrefix = "/metadata/annotations/" + patch2.EscapePointerToken(LabelPrefix)
+
+func dropObjectSetOps(patch []byte) ([]byte, error) {
+	var ops []patch2.Operation
+	if err := json.Unmarshal(patch, &ops); err != nil {
+		return nil, err
+	}
+
+	kept := make([]patch2.Operation, 0, len(ops))
+	for _, op := range ops {
+		if !strings.HasPrefix(op.Path, objectSetOpPrefix) {
+			kept = append(kept, op)
+		}
+	}
+
+	if len(kept) == 0 {
+		return []byte("[]"), nil
+	}
+
+	return json.Marshal(kept)
+}
+
+func applyPatch(gvk schema.GroupVersionKind, reconciler Reconciler, patcher Patcher, debugID string, ignoreOriginal, nullSafe bool, oldObject, newObject runtime.Object, diffPatches [][]byte) (bool, error) {
 	oldMetadata, err := meta.Accessor(oldObject)
 	if err != nil {
 		return false, err
@@ -221,6 +252,25 @@ func applyPatch(gvk schema.GroupVersionKind, reconciler Reconciler, patcher Patc
 
 	if string(patch) == "{}" {
 		return false, nil
+	}
+
+	if nullSafe && patchType == types.MergePatchType {
+		// doPatch generates the merge patch from the ignore-stripped documents,
+		// so the translation has to resolve nulls against those same ones or a
+		// wholesale subtree reinstates a field an ignore patch took out.
+		_, strippedModified, strippedCurrent, err := stripIgnores(nil, modified, current, diffPatches)
+		if err != nil {
+			return false, err
+		}
+
+		patch, err = patch2.CreateJSONPatchFromMergePatch(patch, strippedModified, strippedCurrent)
+		if err != nil {
+			return false, fmt.Errorf("json patch generation: %w", err)
+		}
+		patchType = types.JSONPatchType
+		if string(patch) == "[]" {
+			return false, nil
+		}
 	}
 
 	logrus.Debugf("DesiredSet - Patch %s %s/%s for %s -- [PATCH:%s, ORIGINAL:%s, MODIFIED:%s, CURRENT:%s]", gvk, oldMetadata.GetNamespace(), oldMetadata.GetName(), debugID, patch, original, modified, current)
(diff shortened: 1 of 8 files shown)


pkg/apply/desiredset_compare.go
```diff
@@ -185,7 +193,67 @@ func sanitizePatch(patch []byte, removeObjectSetAnnotation bool) ([]byte, error)
 	return json.Marshal(data)
 }
 
-func applyPatch(gvk schema.GroupVersionKind, reconciler Reconciler, patcher Patcher, debugID string, ignoreOriginal bool, oldObject, newObject runtime.Object, diffPatches [][]byte) (bool, error) {
+// annotationsPath is the JSON Pointer of the annotation map, and
+// objectSetOpPrefix the prefix of the operations that carry apply's own
+// bookkeeping annotations.
+const annotationsPath = "/metadata/annotations"
+
+var objectSetOpPrefix = annotationsPath + "/" + patch2.EscapePointerToken(LabelPrefix)
```
jakefhyde: These should likely goin the `const` and `var` blocks at the top of the file

pkg/apply/desiredset_compare.go
```diff
@@ -185,7 +193,67 @@ func sanitizePatch(patch []byte, removeObjectSetAnnotation bool) ([]byte, error)
...
+
+func stripObjectSetAnnotations(value ejson.RawMessage) (ejson.RawMessage, bool, error) {
+	annotations := map[string]interface{}{}
```
jakefhyde:
```suggestion
	annotations := map[string]any{}
```

pkg/apply/desiredset_compare.go
```diff
@@ -185,7 +193,67 @@ func sanitizePatch(patch []byte, removeObjectSetAnnotation bool) ([]byte, error)
...
 
-func applyPatch(gvk schema.GroupVersionKind, reconciler Reconciler, patcher Patcher, debugID string, ignoreOriginal bool, oldObject, newObject runtime.Object, diffPatches [][]byte) (bool, error) {
+// annotationsPath is the JSON Pointer of the annotation map, and
+// objectSetOpPrefix the prefix of the operations that carry apply's own
+// bookkeeping annotations.
+const annotationsPath = "/metadata/annotations"
+
+var objectSetOpPrefix = annotationsPath + "/" + patch2.EscapePointerToken(LabelPrefix)
+
+func stripObjectSetAnnotations(value ejson.RawMessage) (ejson.RawMessage, bool, error) {
```
jakefhyde, line 202: Can we get comments on any new functions added

pkg/patch/jsonpatch.go
jakefhyde, line 1: `interface{}` -> `any`

jakefhyde (changes requested):
Mostly nits.

crobby (approved):
I think this looks good.  The tests are pretty comprehensive.

marytlf:
Thank you for reviewing it @crobby
@jakefhyde I addressed the nits you suggested and also re did all my manual tests to make sure everything was fine, I think we are good :) thank you for reviewing it

PR #752, opened 2026-09-19
fix(generic): return error from ControllerFactory instead of calling utilruntime.Must

AruneshDwivedi:
Replace utilruntime.Must call with proper error handling. utilruntime.Must panics on error, making recovery impossible. Return the error from ControllerFactory so callers can handle failures gracefully.

```release-notes
[BUGFIX] generic: return error from ControllerFactory instead of calling utilruntime.Must
```


brandond (changes requested):
1. use of Must indicates an execution path where failure is unexpected and cannot be handled normally
2. the requested change STILL doesn't return the error, it just uses logrus to print the error with some context - but logrus.Fatalf still causes the process to exit.
3. the panic from Must can at least be caught via `recover`, unlike the `os.Exit` that logrus executes on `Fatal`

in short, the pr doesn't seem to do what you say it does, and seems like it might actually make things worse.

